PR #5, opened 2015-01-22
Auto-leveling and probe file load/saving fixes

mandrav:
Hi,
this pull request fixes two issues:

1. While auto-leveling, the current code ignores non-movement lines, e.g. F200, M3/M5, etc. This fixes it.
2. The default bCNC.ini, which is used in new installations, does not contain a 'probe' option in the 'File' section which raises an exception in profile file saving/loading. This is also fixed by catching the exception and using a default value in this case. Although bCNC.ini should be updated also to contain this File->probe option so no exception is raised in new installations :)

Thanks for this great tool :)
:+1:

### DIFF
--- a/CNC.py
+++ b/CNC.py
@@ -807,7 +807,9 @@ class CNC:
 				self.processPath(cmds)
 				xyz = self.motionPath(True)
 				if not xyz:
-					newlines.append(None)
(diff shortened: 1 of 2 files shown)


vlachoudis:
Thank you!

PR #11, opened 2015-04-05
Don't convert numbers after G, T, M or N commands.

toomuchcookies:
Grbl sends a "command unsupported" error for a G1. or G0. command. This happened, when bCNC 'moved' the gcode.

Signed-off-by: Omar Abo-Namous <kontakt@toomuchcookies.net>

### DIFF
--- a/CNC.py
+++ b/CNC.py
@@ -1442,6 +1442,9 @@ class GCode:
 			new.clear()
 			for cmd in cmds:
 				c = cmd[0].upper()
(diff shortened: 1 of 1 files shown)


vlachoudis:
Thank you Omar, this error was corrected in the last upload. It appeared when I modified the formatting to allow more precision for the floating point numbers when doing profiling operations. Can you test the last version to check?

toomuchcookies:
This seems to work as well.

PR #26, opened 2015-05-07
Show progress in HH:MM:SS

Effer:
Show current and estimated time in hours,minutes and second intead of
only seconds

### DIFF
--- a/lib/tkExtra.py
+++ b/lib/tkExtra.py
@@ -552,7 +552,7 @@ class ProgressBar(Canvas):
 			tot = dt/p*(self.high-self.low)
 		else:
 			tot = 0.0
-		self.tmsg = "[%.1fs : %.1fs]"%(dt,tot)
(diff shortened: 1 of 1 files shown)


vlachoudis:
Thanks. However the minutes as you calculate them are wrong, they contain the hours inside. E.g. for an 1:30:00 run will be displayed as 1:90:00.
I've made the addition to display hours/minutes/seconds using the divmod

Effer:
Not tested enough. It was more a way to understand how pull request works.
Thanks for the insight.
.

Filippo

----- Messaggio originale -----
Da: "Vasilis Vlachoudis" <notifications@github.com>
Inviato: ‎07/‎05/‎2015 21:03
A: "vlachoudis/bCNC" <bCNC@noreply.github.com>
Cc: "Filippo" <f.rivato@gmail.com>
Oggetto: Re: [bCNC] Show progress in HH:MM:SS (#26)

Thanks. However the minutes as you calculate them are wrong, they contain the hours inside. E.g. for an 1:30:00 run will be displayed as 1:90:00.
I've made the addition to display hours/minutes/seconds using the divmod

PR #33, opened 2015-05-12
Remove unnecessary open of tab Terminal

Effer:
in case of enabling "Check GCode".
The status "check" is always visible in the status display.


Effer:
I am touched, it's my first merge!
I'm going to celebrate with a coffee!

PR #35, opened 2015-05-13
Added a screenshot for wiki

Effer:
@vlachoudis  is there another way to add some picture to the wiki? I saw you use this way to show bCNC snapshot (that it's quite outdated by the way) but it seems to mix doc and software for me.
If you prefer in another way let me know.


vlachoudis:
I am afraid don't know how. I found this way but is not very nice. Maybe on the github help pages there is a better way.

Effer:
A dirty trick could be to open an issue and attach an image there, then reference it in the the documentation. But I'll look to other way.

Effer:
I found this is a more clean solution : https://publicobject.com/2014/12/31/posting-images-on-githubs-wiki/
I've tested in my wiki fork https://github.com/Effer/bCNC/wiki/Connecting It works good. Unfortunately GitHub doen't support pull request for wiki.
For the moment I'll avoid adding images to wiki.

chamnit:
FWIW, I simply upload pictures to the main github repo under some images folder or host it on my Dropbox account. I just have the Wiki link to the pictures with the embed image markdown.

vlachoudis:
It should be ok, as long as they are out of the trunk folder

Effer:
Sorry @vlachoudis but I don't get how you prefer the picture been added to wiki. I'd prefer a solution that require less overload possible for you but I'd like to avoid storing it to some external website with the risk loose them in future.

vlachoudis:
@Effer its better to be on the repository, but I meant on a folder outside the trunk. So whoever downloads the code he doesn't download as well all the screenshots. I don't know if git permits us to create a screenshot folder outside the trunk?

chamnit:
You could make something like a 'doc' branch and place everything there. There is also a special branch called 'gh-pages' that will serve a webpage.

Effer:
May be @vlachoudis could add the 'doc' branch or add another repository like @chamnit  is doing for grbl builds?

vlachoudis:
@Effer I've created a doc branch, so we could push the documentation and screenshots there

PR #41, opened 2015-05-16
Updated pendant HTML file

jarretluft:
Hey Vasilis,
I've made some changes to the pendant interface to hopefully make it more compatible with a wider range of browsers (desktop and mobile) as well as simplify some of the coding.
Changes are as follows:
* Updated the web server in CNCPendant.py to return different header information to the browser depending on whether the file extension is CSS, JS, or HTML
* Refactored all javascript to use the jQuery framework. This simplifies the ajax request code, and builds in fixes for browser specific 'anomalies'. 
* Added pure.css framework to manage page layout across different browsers and screen sizes.  This allowed me to remove all the tables in the original and use DIVs instead, with the CSS classes handling the positioning of elements in a grid.
* Added functionality for the coolant drop-down as well as spindle
* Added checks for spindle and coolant state in the getState() function to update those drop downs alongside the WCS whenever new info is received in the $G array.
* Added $G status requests in a few places (after a WCS change, or spindle/coolant change, to ensure the bCNC state variable is as up-to-date as possible.  I also added $G and $# requests on loading the pendant to ensure that the variables in bCNC are as up-to-date as possible when updating all the html drop down boxes on the pendant interface.
* Added the FastClick javascript library to improve the speed of the pendant on mobile devices that may be affected by the 300ms lag issue (IOS, android)

These changes appear to have fixed some of the problems reported in one of the recent issue requests, and in my testing I have not experienced any further issues with the button action occurring twice on click.

Please give this a try and merge in the changes if you think they are helpful to the project.
Cheers,
-J.


chamnit:
@vlachoudis : I've tested this new virtual pendant and can verify that it fixes all of the issues I ran into. No triple clicking in Safari and significantly better responsiveness across all of the browsers I tested. 

The only minor new issue was related to the 'Pause' job button. I had to press it twice to initiate the pause and resume. It might be related to the bCNC internal pause function, rather than this web interface.

I'm pretty excited, since this is a great framework for anyone to modify and make their own custom virtual pendants. If I have time, I may even look into a browser-based VNC viewer. Not sure if it'll work well or not, but it'll be interesting to see if it'll work on a tablet. VNC is great because it's just a mirror of the main GUI. If it crashes or loses a connection, no big deal.

.gitignore
```diff
@@ -0,0 +1,2 @@
+*.pyc
```
chamnit: @vlachoudis : I noticed that Jarret removes the .pyc files from getting sync'd into the github repo. Do you add these for performance purposes? Or can they be removed?

jarretluft:
I removed those because they were created at compile time on my system, not part of the original clone of the bcnc branch, and I didn't want them included back in my pull request as new files.

vlachoudis:
Thanks @jarretluft. I just returned home, since I am new to github, how can I download the modified version as a whole package to test?

jarretluft:
My fork is at:
https://github.com/jarretluft/bCNC

You can clone the full repo and test.

vlachoudis:
@jarretluft @chamnit Ok, thanks I'm learning how the github works.

Just profiting from you and your web experience.

1/ How can I make the style of the Command Line (MDI) to resize to the width of the screen?

2/ I wanted at some moment to make a virtual rotating dial wheel is it possible on javascript?

chamnit:
I'm not a web programmer either, but a researcher like you. Matlab and C are my best friends. (some Python too). I view this virtual pendant code as a first step. Validate that bCNC's backend is capable of serving a browser-based interface. If so, it'd be possible to do just about anything, like a dial (might be tricky). Or run bCNC as a command-line utility without the GUI and only run it with the browser like a stripped-down Chilipeppr. Or users can make customized html pendant files for their personal workflows. 

As far as I can tell, the virtual pendant works really well. There are still minor lag issues (likely not caused by the JS/html but the python backend), but the basis for this is all there.

1bigpig:
@vlachoudis I am no web programmer but there is a cool web UI dial on such projects as Rune Audio (volumn and track length round dials).  I do not know what language they are programmed in, but it might be worth a look-- they claim "free and open sourced".
http://www.runeaudio.com/

Bruce

vlachoudis:
@chamnit  I have the feeling that the lag is caused by the opening/closing communication of the javascript. On every click it has to open a new connection to the web server, python has to spawn a new thread, read the data, parse it and return the json information string. If this is the problem I could see two possible solution to test:
1. Kill the multi threading on the web server and permit only one connection only.
2. Maybe @jarretluft  can help on that,  if we could open a tcp port through Js, and let it open as a bi-directional communication js-python with a simple protocol.
@1bigpig thanks for the information, I will give a look on it.

jarretluft:
@vlachoudis making use of a tcp websocket for communication between python and javascript might solve the lag, this would allow bidirectional communication so instead of having to request status changes every second from python through ajax URL requests, those updates could be pushed back to the pendant whenever a status changes on the server side.  It would also allow a permanent connection between the javascript browser page and the python application until the browser is closed or the pendant server is stopped.
I can play with this idea and see if there is a workable solution that doesn't add too much overhead to the python program.  One concern is that it may add another dependency to the bCNC install if a separate python websocket library is needed.

chamnit:
@vlachoudis : You're probably right about the lag. Although, it's really not too bad. It's noticeable on a smartphone or tablet, but not on a laptop web browser on a local network. 

@jarretluft just beat me to the web sockets suggestion, but doing a quick web search there are a few libraries that provide web socket support for Python [link](http://www.tornadoweb.org/en/stable/).

jarretluft:
@vlachoudis @chamnit I rewrote the pendant server and client to use the websocket protocol last night, all seems to be working but I haven't tested the response speed on mobile yet to see if it is improved.
This version just takes the existing communication format that was being sent with the GET requests and translated it to go through websocket, but now that I have a working concept I could move it to a push update system where the python back-end pushes changes to all connected clients, and the client doesn't have to continually request updates every second.

Also, I am using a simple [websocket server python script](https://github.com/Pithikos/python-websocket-server) that I included directly in the bCNC lib folder to avoid creating another system dependency like using tornado or autobahn would have.

Still some work to be done before it's worthy of a pull request to the main branch, but the functional proof of concept is here:  https://github.com/jarretluft/bCNC
(my python code may be a little sloppy - still learning)

chamnit:
@jarretluft. I remember to test it before leaving home this morning. It subjectively works better on an iPhone. But I noticed that it doesn't take multiple button pushes very well. For example, with jogging, you'd often push the jog button several times. It seems to only execute a new motion and button press after the current one has completed. Not a big deal, but can't remember if the pre-web socket version would queue multiple button pushes or not. I'll check later when I get back to my shop.

vlachoudis:
@jarretluft it seems that the websockets are not a standard package to python. In this case it would be nice to have a check on the web server if the websocket import is existing to serve with it otherwise with the standard http protocol. So the users that don't use the pendant, not to go in the hassle of installing an additional package.

chamnit:
@vlachoudis : AFAIK, the web socket server that jarret selected doesn't have external dependencies. It's a simple python file that uses the standard SocketServer. It worked on both my Mac and RPi without needing to install anything additional.

jarretluft:
There is no additional requirements to run this, the websocket python library used is included in the bCNC /lib folder.  I just linked to the original source above for purposes of attribution.  I moved the necessary class into: https://github.com/jarretluft/bCNC/blob/master/lib/websocket_server.py

vlachoudis:
@jarretluft brilliant. I like it this way.

chamnit:
@jarretluft @vlachoudis : I nearly forgot about this. @Protoneer has been working on a Raspberry Pi CNC controller [[link](https://github.com/Protoneer/Raspberry-Pi-CNC)], based on GrblWeb with a Python server coincidentally. He doesn't use the dependency-free web socket_server that Jarret has, but it looks like Protoneer has a lot of things figured out. Templates for different interfaces, JSCut integration, and a g-code viewer. Maybe this is the next step?

chamnit:
@jarretluft @vlachoudis : I just tested this on my system. There isn't much noticeable difference between the web socket and the non-websocket(updated) version of the pendant on an iPhone. The button presses are all limited to about 3 per second. They all worked great. 

But, there was a minor difference between controlling the pendant from my Macbook Pro's Safari. The mouse clicks were slightly more responsive with the web socket implementation. Again, both seemed to work great.

Issue #45, opened 2015-05-16
Mouse position display in inches


Scott216:
When I move the mouse around in the visualizer bCNC displays the mouse position at the bottom of the screen, but it's in mm.  I think if I'm using inches, the mouse position should be inches also.

vlachoudis:
@Scott216 the visualizer shows the units in what you choose in the settings. The default one is mm but you can change it through the tabs "Tools" -> CNC -> Units
Then restart the program. I believe the suggestion of @Effer  is good to place also the units on the status bar.

Scott216:
If I change the units to inches, should I also convert all the other values in the CNC section, or will bCNC do that for me?

vlachoudis:
That's a good one! I was not sure what it will do, but testing shows that if you change to "inches" and close the program, then next time you open it, everything will be converted to inches.
Which proves to be a problem the first time someone changes the units and he thinks
that he can enter the units in inches, the program will still think in mm.
I have to correct it.

The program internally  and in .ini it writes everything in mm (I am thinking in metric :),
but it converts and displays to what the user selects.

PR #46, opened 2015-05-16
Mouse position display in inches #45

Effer:
Make some test and  it work in inch now.
I've not look in deep if was better to change the output from canvas2xyz(). That is used in many others parts that works already in well.


Effer:
I've notice a small defect in the two label on the bottom of bCNC. When moving coordinates with mouse the two widgets resize himself in an erratic way. You .pack them, but maybe using .grid could avoid the flicker.

vlachoudis:
@Effer I would like to merge part of the pull requests, but the system doesn't allow me. Do you know how I can select which commits I could merge

Effer:
I'm new to git too, so take with caution

**git cherry-pick "hash commit number"**

Extracted from:
http://markosullivan.ca/how-to-handle-a-pull-request-from-github/

You should clone my report and the check out it. Go to your branch and cherry-pick the interessted commit. If in doubt copy & paste  what you need ;)

Effer:
A passage was missing, you also had to add my repository as remote and fetch it. From then you can cherry-pick the commit you need.

http://stackoverflow.com/questions/5120038/is-it-possible-to-cherry-pick-a-commit-from-another-git-repository

PR #77, opened 2015-07-09
Changed probing pattern so it doesn't travel to starting X pos for each row

mandrav:
Probing would go from x-begin to x-end, change row, move back to x-begin and start again for the next row.
With this change, when it reaches x-end, changes row and starts probing backwards in the other direction.
Less travel moves, better for us with slower CNCs :).


Effer:
That's a good point.

vlachoudis:
Indeed this is very useful

PR #79, opened 2015-07-09
Separated goto-zero button into two buttons: one for XY (existing button) and one for Z (new button)

mandrav:
Well, as the title says :)
I find I need to return to home on XY quite often (but not on Z0) so I added a new button for separately going to Z0 while the existing button will only goto X0Y0.

### DIFF
--- a/bCNC.py
+++ b/bCNC.py
@@ -509,7 +509,15 @@ class Application(Toplevel):
 
 		# ---
 		row += 1
-		col = 1
+		col = 0
+		b = Button(lframe, text=Unicode.LARGE_CIRCLE,
+					width=width, height=height,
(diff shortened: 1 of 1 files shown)


vlachoudis:
Thanks. for the moment I put this on hold, since it conflicts with the other pull request from @onekk  #73 and in any case I am re-writing the interface.
However since you pointed out, there was a bug in the go2origin button, because it was recently overridden by a UserButton and the command is defined inside the bCNC.ini. By right-clicking the button you can change the command, and as well you can add the go2originZ on any user button.

mandrav:
Nice :)
Feel free to ignore this too then.

Yiannis.

mandrav:
Out of date

PR #80, opened 2015-07-09
Serial port opening a little more robust

mandrav:
Sometimes I have trouble connecting to the serial port.
The only way I 've found to make this work again is to open the arduino ide serial monitor. This somehow resets correctly the port and I 'm then able to connect again with bCNC.

After some source digging and googling, I made this port opening a little more robust. It 's been working fine for me now for more than a month so I don't think any of the extra commands have any side effects.


vlachoudis:
Thanks

PR #94, opened 2015-08-25
Feed Override Test

Effer:
Just a proposal to develop something for change feed override, waiting a grbl realase that manage overrides in a better way.


Effer:
Something gone wrong with my pull request, sorry.

PR #98, opened 2015-08-29
Feed Override

Effer:
Feed Override Test, 
I've made some test to implements an usefull function for override feed during line feed to grbl.
I know that this approach is far from optimized, but this feature is realy a must for test materials with cnc.
I hope that grbl 1.0 will finaly implements feed overrrides.

### DIFF
--- a/bCNC.py
+++ b/bCNC.py
@@ -796,6 +796,19 @@ class Application(Toplevel):
 				command=self.stopRun)
 		tkExtra.Balloon.set(b, "Stop running program")
 		b.pack(side=LEFT,expand=YES,fill=X)
+		
+		# Override
+		self.override = IntVar()
+		self.override.set(100)
+
+		b = Scale(f,
+				command=self.overrideControl,
+				showvalue=True,
+				orient=HORIZONTAL,
+				from_=25,
+				to_=200)
+		b.pack(side=BOTTOM, expand=YES, fill=X)
+		tkExtra.Balloon.set(b, "Set Override")
 
 		self.progress = tkExtra.ProgressBar(lframe, height=24)
 		self.progress.pack(fill=X)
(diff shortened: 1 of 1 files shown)


vlachoudis:
Thanks @Effer. One modification that can make the feed rate scanning more efficient is
to use regular expressions like
`FEEDPAT = re.compile(r"^(.*)[fF](\d+\.?\d+)(.*)$")`   # at the top of the program

`pat = FEEDPAT.match(tosend)`
`if pat is not None:`
`   fi = float(pat.group(1)) * self.override.get() / 100.0`
`   tosend = "%sf%g%s"%(pat.group(0), fi, pat.group(2))`

the only thing is that it will override the feed only when it is defined and not for the lines that it is not existing.

Effer:
Regular expression is surely a cleaner suggestion.
To have a quicker response over override changes I could append a Feed command with override value to the streamer. This should happen only the first line after an override change.
All other line will be affected as soon as they are send.

Do you believe this will be a useful feature?
From my test,  feed changes affects grbl a bit too late. Mostly depends from the type of created gcode. If grbl has buffered a lot of work, the feed will change with a big late.
But when testing new material or tools it reveals really handy.

chamnit:
@Effer : This would be a temporary fix, as I have overrides working in unpublished developmental builds. It's more of a question of getting it to fit on a 328p and finding the time to finish it. If anything, it'll run on a mega2560 just fine.

Effer:
@chamnit  : that feature is on the top of my wish list, but I found an old post (2013) about the implementation, and I thought It couldn't be implemented.
Hope you can fit it on a 328p for the joy of many so this fix will be useless.
Thanks for your great work.

chamnit:
@Effer : Now that I think about it. If feed overrides doesn't fit on the 328p, I think this feature would still be useful for people who are stuck with integrated 328p stepper driver boards.

PR #99, opened 2015-09-07
Fix python path for plugin directory on windows os

Effer:
For the "poor" windows user ;)


vlachoudis:
Thanks

Effer:
You're welcome

PR #100, opened 2015-09-07
Pendant upload file

Effer:
I find this a missing feature in the pendant interface, an idea for an easy way to upload file from another  terminal.
I suggested it a while ago, but at that time it seemed that the web interface was completely rewritten. Now some time had passed and maybe it can find a place in this eval branch.
It's still quite raw and untested, comments,suggestions (or rejections) are welcome.

### DIFF
--- a/Pendant.py
+++ b/Pendant.py
@@ -13,6 +13,7 @@ import sys
 import json
 import threading
 import urllib
+import re
 
 from CNC import CNC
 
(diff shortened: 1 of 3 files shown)


vlachoudis:
I've merged this request, but I had to restore the "monitorSerial" function

vlachoudis:
@Effer I don't see any button in the html to upload the file, there is only a browse

vlachoudis:
@Effer forget my previous comment, there is no need it automatically uploads.
It works, however it uploads a file from the server disk, I though you are sending
the whole file from the client side.
You didn't need to create a new variable, but just to push a command in the pandant stack
as "load <filename>"

Effer:
If bCnc is running in a local network, it can recive file from any client (cam workstation), store it in a folder (user subfolder now), load it ready to run.

My first try was with a "load" command in the stack. Can't remember why but didn't work. Maybe some issues with method parameters?
I'll try again later.

Filippo

----- Messaggio originale -----
Da: "Vasilis Vlachoudis" <notifications@github.com>
Inviato: ‎07/‎09/‎2015 18:07
A: "vlachoudis/bCNC" <bCNC@noreply.github.com>
Cc: "Filippo" <f.rivato@gmail.com>
Oggetto: Re: [bCNC] Pendant upload file (#100)

@Effer forget my previous comment, there is no need it automatically uploads.
It works, however it uploads a file from the server disk, I though you are sending
the whole file from the client side.
You didn't need to create a new variable, but just to push a command in the pandant stack
as "load "

Effer:
@vlachoudis  Made a small test for remove the new variable:
I've use httpd.app.queue.put("load \""+fMsg+"\"\n") but there are some issue with space path (at least in windows), they are truncated even with double quotes.

Another question, in the sender class there area couple of queues, one for commands to grbl and the other for pendant commands. Is the pendant queue intend to be used for the web interface? It seems not used anyway.

vlachoudis:
I have to check where is the space stripping is happening.

For the moment the Sender and interface are not fully separated. I will get there soon :)
So the pendant queue is consumed inside the monitorSerial thread of the interface

PR #112, opened 2015-09-27
Spirograph plugin

Effer:
This is a simple plugin to generate spirograph path (roulette curves) https://en.wikipedia.org/wiki/Spirograph.

spirograph

Spins number are self calculated, so is simply a metter of input Internal, External radius, offset and desired deep. All works based on materials and bit installed.

Usefull to generetate some geometric shape for decoration:

wp_20150927_21_22_52_pro

Other plugins help me a lot as base for undestand the mechanism. But would be usefull to have some more info in the wiki.


vlachoudis:
Cool!

vlachoudis:
The wiki needs heavy updating, as well to add a pointer in the grbl wiki for bCNC

vlachoudis:
I've just tried the default values and it seems it didn't work. I've got a nice circle multiple times :)

Effer:
Sorry @vlachoudis, some points was missing, I've just PR a fix.
And Yes, wiki need a lot of work, but it's not so easy stay up with all yours updates, simply you code too fast. ;)

PR #117, opened 2015-10-01
Add stepover for End Mill

Effer:
Is this enough to add StepOver parameter to End Mill?


vlachoudis:
Almost. It needs also to propagate in the "def update(" the value to CNC. I will do it

vlachoudis:
By mistake I've close it before merging it. I'll do manually :)

Effer:
No problem, I'll use it for pocketing tool.

vlachoudis:
Ok done

PR #120, opened 2015-10-02
Cleaned up icons

mandrav:
I edited bCNC.png and bCNC.xbm files to make them a tad cleaner.
bCNC.png is also now better when viewed in dark backgrounds because I added a light outline to the dark icon.


Effer:
That was a change that I also kept in the drawer for some time.
@vlachoudis doesn't like to play with graphics ;)

mandrav:
@Effer , as long as @vlachoudis likes to play with code, I 'm fine with it ;)

PR #124, opened 2015-10-09
Flatten Tool

Effer:
Here my attempt to add a pocket generator.
Logic behind: create a Raster (zig/zag) or Offset (spyral) path that alway feed the tool with desired stepover. For Raster approach is possible to enable a border pass.
Cut direction Climb or Conventional (for Raster only for the border)
The pocket area can be define in width and height (obviously) and x,y offset
Still missing:
- tests from experienced cnc worker !!!
- optimize work direction x/y to reduce passes or change it on user choice
- implements some @vlachoudis  suggestions like get default area to work from bCNC

flatten

Results on material seems good to me, comparable with others cam output. Still I'm not completely sure if some choice are the best. 
Beta tester are welcome, don't blame me if your bits broke!


vlachoudis:
nice!

carlosgs:
Very convenient, thanks Filippo!!!

Effer:
Thanks!
bCNC is very fun!

PR #142, opened 2015-10-25
Save probe data as STL file.

Effer:
This time not a plugin ;)
It's more a integration on what already exist for probed data.
Probing with high details take a lot of time, but result are nice.

Here the original:
scanoriginal

And here the exported result as STL file:
scanrecreated png

I've chose stl because seems really appreciated by 3d makers.

### DIFF
--- a/CNC.py
+++ b/CNC.py
@@ -17,6 +17,7 @@ import undo
 import Unicode
 
 from dxf   import DXF
+from stl   import Binary_STL_Writer
 from bpath import Path, Segment
 from bmath import *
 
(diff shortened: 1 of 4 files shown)


vlachoudis:
You could have used the Vector class from bmath.py inside the stl library.

Effer:
I'm a bit lazy...

2015-10-25 19:32 GMT+01:00 Vasilis Vlachoudis <notifications@github.com>:

> You could have used the Vector class from bmath.py inside the stl library.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/vlachoudis/bCNC/pull/142#issuecomment-150952098>.
>



-- 
Filippo

Harvie:
I usualy get much better results when i only use pointcloud from 3D scan and apply poisson surface reconstruction in meshlab to generate STL. the STL generated directly by bCNC is not very nice.

PR #145, opened 2015-11-10
Driller plugin

Effer:
Here a plugin to create holes along selected block of code.
The plugin avoid G0 rapid movement and all movement with same Z (it make little sense to pile hole one over the other). Then it calculate the length of the path and make holes at fixed distance. Remaining space is equally distributed at begin and end of path.
Elaborated block are disabled after the process.

Here a test importing a DXF:
driller
  
and the result:
wp_20151110_20_52_39_pro

It works also with G2/G3 arcs, here you can notice that Z movement doesn't present stacked holes :
drillertest

There is plenty of space for improvements.


vlachoudis:
Cool! You are now officially the plugin-master :)

Effer:
Really proud! : D

dgoadby:
This would be useful for chain drilling where you are going to remove a piece of metal from a sheet. Drills are cheaper than milling cutters so chain drilling is a good way to remove large lumps of material. Doing chain drilling by hand takes a long time so this plug-in is really a stonking idea - well done!

Effer:
Thanks @dgoadby.
I wrote the plug-in without a clear idea of ​​its real use. Fortunately I am finding many useful suggestions!

chamnit:
@Effer : Does this do [peck drilling](http://linuxcnc.org/docs/html/gcode/g-code.html#_g83_peck_drilling_cycle)? This would be very handy for deep and large holes drilled in metal.

Effer:
@chamnit it make **peck** and **dwell**, but doesn't use  G83 syntax, I've preferred to expand all gcode using G0 and G1 commands. Implementations should be similar:
make the hole at peck depth (if present), stay for dwell time (if present) retract fast to safe. Repeat increasing depth of peck parameter until target depth is reached.

PR #152, opened 2015-11-17
Step over as pure number?

Effer:
Hi @vlachoudis isnt' this to be consider as pure number and so not to be converted from mm/inch?


vlachoudis:
Thanks

PR #158, opened 2015-11-19
Heightmap plugin

Effer:
A new plugin ready for weekend.
It's a porting of [image2gcode](http://www.linuxcnc.org/docs/2.4/html/gui_image-to-gcode.html) from LinuxCNC and another python implementation from scorch: [dmap2gcode] (http://www.scorchworks.com/Dmap2gcode/dmap2gcode.html) 
It take an image and create a height pocketing using darkest area as deeper and lighter area as higher. Here some fast test (bad quality):

From:
dame
To:
wp_20151119_15_02_05_pro

or from this:
bcnc

to this:

screenshot 2015-11-19 19 17 10

and this:

wp_20151119_19_18_18_pro

It needs PIL and try to use numpy if installed.

Documentation is on the way (next days).
Enjoy!


vlachoudis:
What will be the next plugin?   :)

Effer:
Still not sure.. just a couple of ideas :)

chamnit:
@Effer : Agreed. What's the next plugin?! :) 

One question though. It looks like you do everything in one pass and don't consider depth of cut, which can be problematic and lead to broken bits. Is it possible to do some minor image processing and do multiple passes at a maximum depth of cut? I suppose I mean, take the image, temporarily filter out all color to the color of the first depth of pass, generate g-code for the filtered image. Reprocess the image to do the next max depth of cut. Repeat until you get to the final depth desired.

I know this can get a lot more complicated than this, like avoiding recutting areas between passes, but I thought I would bring this up, since this plug-in is AWESOME.

Effer:
@chamnit It's already there, for screenshoots I've choose to use only one pass to be clear.
It takes parameters from Material --> Depth increment
It also avoid to overcut already processed path

What I could add (is there but hidden for now) is the use of multiple tools, one for rough till an offset and the other for finish. That is not so easy to integrate with bCNC.

chamnit:
@Effer : Awesome. Do you do rapids in the areas that already have been cut? For example, the second depth of cut layer avoids recutting the first layer and just rapids out and over that section?

If not, maybe the plugin could do something where if an image has and specific color or transparency, the plugin will perform a rapid over the area.

As for roughing and finishing, I think it's possible to do it with bCNC. You just to process the image with different brightness values for the roughing pass. Althoguh I'm probably missing something, as I don't know the specific workings of bCNC or your plug-in.

carlosgs:
Great work, an easy to use open-source tool for this was lacking for a long time. Simply amazing to have it integrated into bCNC. Thanks so much Filippo!!!!

Effer:
@chamnit @carlosgs thanks for compliments!
As said most of the work is based in image2gcode and they do a great job. I've verify that most of the code is optimize to avoid work on the same area if not need. 
The whole solution is really well structured and I've exposed only the most relevant part integrating it with bCNC.
@vlachoudis I think this the first plugin to use the **shape** of the tool, indeed it take care of 3 different type of end mill (Flat - V shape - Ball) and generate different path from different geometry. I've grouped the existing types to fit this ones. The code behind is quite complex and take some time to complete. During this the GUI freeze (at least on my ten years old pc..)

I'll make more test and collect some feedback before adding also the **offset** parameter that will stop the  roughing path before reaching the target depth. In this way it would be possible to prepare the material with a bigger EndMill and then finish it with a smaller tool. 
I've just to figure out the most user friendly way to do this, avoiding to duplicate material with different **depth increment**.

carlosgs:
"avoiding to duplicate material with different*depth** increment*."

Maybe it is possible to simply have a checkbox that says "single pass" that
avoids pocketing when checked?
It'd be similar to the "pocket" option in the bowl plugin.

Effer:
That was exactly my first tough Carlos. It seems the best way and without huge modification.

1bigpig:
@vlachoudis <https://github.com/vlachoudis> I have about 5 plug-in ideas,
but I am in the middle of "Introducing Python" book, so it will be a little
while before I/they are ready.

Here are a few of my ideas:
1.  tab generator (little uncut material--tabs--to hold cut through
material in place)
2.  a few simple shape generators.  Stars and other simple shapes based on
size and number of points
3. a fun and sometimes useful Hilbert curve generator...good for background
patterns on carved pieces.
4. maybe a bolt hole circle calculator
5.  some other stuff...I am full of ideas

I do have one question.  I have not looked at the bCNC code nor have I had
a chance to look at any of the plug ins yet.  Does the programmer have
access to selected geometry?IE can I make the tabs generator interactive
based on which geometry is selected or which geometry the user clicks on in
the window?

Also, is it bad etiquette to bring up an interactive window in the plug-in
window?  IE if I want to draw a star or other shape interactively as the
values change, before the object is "pasted" into the geometry list.

Back to the book...
Bruce

vlachoudis:
@1bigpig new ideas are always welcome :)
Tabs are partially inside. The interactive editing is missing. Just for testing you can add tabs with the 't' short cut only and change the parameters in the Tools->Tab

You have full access to everything. In the plugin either you introduce your new gcode or you modify the existing.

Nothing is preventing in using additional dialogs/windows inside the plugins.

vlachoudis:
One plugin that I was thinking was to break/split a large gcode into smaller pieces fitting in the workarea

Effer:
@vlachoudis that will be useful for small bed machine.

Also a **nesting** feature will be fit nice. The code behind usually needs high computational power, but even an rough approach could help to optimize waste of material.

onekk:
Effer I've worked on a modified dmap2gcode for grbl you will find at https://github.com/onekk/grbl-i2g, the slicing routine try to use the maximum speed over already machined area, expecially for finishing pass, for now it is stalled as I've bought a proper CAM, but if you find it useful, as fas as I know Scorch use the cutting speed to machine the already machined area when you set the finishing pass, and it is a waste of time. See the comment in the code for the routines, see if you want also my eurocam attempt that use Anders Wallin routine opencamlib to find the 3d contour taking in account the tool shape (slightly modified the Pyhton code to make them compatible with the new graphincs interface).

Regards

Carlo D.

Effer:
Just for the historians, the option for the finish pass has been added as suggested from @CarlosGS . Better late then never. I totally forgot.
Here a test: rough pass with flat end_mill 3.175,  multiple pass each at 1.5 depth. Then finishing with a sphere end_mill 1.6. Single pass.

image


wp_20160503_22_02_04_pro

Enjoy

CarlosGS:
Neat, thank you!
Very nice hand btw :D
On 5 May 2016 21:32, "Filippo" <notifications@github.com> wrote:

Just for the historians, the option for the finish pass has been added as
suggested from @CarlosGS <https://github.com/CarlosGS> . Better late then
never. I totally forgot.
Here a test: rough pass with flat end_mill 3.175, multiple pass each at 1.5
depth. Then finishing with a sphere end_mill 1.6. Single pass.

[image: image]
<https://cloud.githubusercontent.com/assets/1007007/15054577/71d2955a-1308-11e6-929b-58f6d7dd640a.png>

[image: wp_20160503_22_02_04_pro]
<https://cloud.githubusercontent.com/assets/1007007/15054471/efb2f308-1307-11e6-9778-0fae8ec19dd2.jpg>

Enjoy

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/vlachoudis/bCNC/pull/158#issuecomment-217253628>

PR #159, opened 2015-11-23
Change invoked shell to sh for wider platform support

justinclift:


justinclift:
Trying out bCNC on OpenBSD 5.8.  Seems ok in first testing, after modifying this launch script to not depend on hard coded bash path. :wink:

PR #172, opened 2015-12-14
Text plugin

Effer:
Here a plugin to generate text outline using True Type Font.
The code is still beta:
- complex glyph are missing (now it return the empty char)
- beizer curve are converted to polylines without consideration to real size (arc accuracy)
- very few test in real world

Main problem was a ttf font library for python to integrate with bCNC. Simply it was missing, so I take code from pyglet (thanks to Alex Holkner ) and ported some JavaScript code to extract glyph vectors.
I'm a novice on both language and probably something is far from optimized.

Here a screenshoot:

textplugin

Then I've added a goodies , if you provide an image as input (and the number of char as width) the plugin generate an Ascii Art rappresentation:

Source:

asciiartmmoriginal

Result:

asciiartmm

This was more a joke and I didn't try to machine anything, maybe someone want to have some test?
Even the integration  of this second feature is not so nice, but splitting in two plugin seems not good idea either.


vlachoudis:
Thanks Filippo! Now you gave me work to create Alignment buttons :)

Effer:
@vlachoudis some time ago I proposed to show measures in the canvas and you spoke about the fact that Text canvas in tkinter doesn't allow to draw rotated text. The ttf library extract vector data so is just a matter of drawing polyline and some math to have rotated text. Could be useful?

vlachoudis:
@Effer Indeed it could be an option. To create a text library based on the ttf as polylines

PR #183, opened 2015-12-28
Adds a "Set WPOS" action to the canvas and keyboard shortcuts for jogging

andyHa:
Sometimes it is helpful to be able to align the machine to an existing distinct point or egde and then update the WPOS according to this known position. This adds an action which lets the user click on a known point to update the WPOS.

It also adds key bindings to "1", "2" and "3" to set three pre defined common step widths (0.1 1 10 by default).


vlachoudis:
Thanks. I've added a protection to avoid changing the step while in any input editing field e.g. MDI/command line or in the editor

Issue #179, opened 2015-12-24
G-Code Optimization?


TilTisback:
Hi,
i know this is not a real issue report here but i don't know where to post it instead.
So here is the thing: is there a way to implement a g-code optimization into bCNC?
There are some of command line based optimizators for the traveling salesman problem out there but it would be very nice to have such a tool inside the program itself. Like load a gcode, hit optimize and see the result directly in the main viewport.
I would really appreciate it!

Effer:
@TilTisback 
something already there, maybe a bit hide:

![image](https://cloud.githubusercontent.com/assets/1007007/11991089/58b5287c-aa16-11e5-8cdc-ff6689ccd31c.png)

From code description:
**Re-arrange using genetic algorithms a set of blocks to minimize rapid movements.**

Is this what you was looking for?

TilTisback:
Ok i haven't seen that but it seems like its not working?
Or maybe i am doing something wrong?
I load my drill file select everything and hit optimize.-->nothing happens
Then i saw it has only one block so i loaded an isolation file with about 209 blocks, select it ,hit optimize and still nothing happening.

The optimization should work for single blocks too right?
Here is what my drillfile looks like. It takes more time to travel than actually to drill.
![drillmess](https://cloud.githubusercontent.com/assets/13772017/11991689/7ce5f814-aa1d-11e5-8490-5b3316479076.JPG)

Effer:
Select all blocks (es: Ctrl+A) and then use the optimize. 
It should work, I've made this test:

Before:
![image](https://cloud.githubusercontent.com/assets/1007007/11992084/f05af160-aa21-11e5-9653-406960c499ab.png)

After:
![image](https://cloud.githubusercontent.com/assets/1007007/11992068/a4a90eaa-aa21-11e5-9e54-cab88d09f603.png)

vlachoudis:
@Effer  is correct you have to select only the blocks of gcode you want to optimize. Maybe I should put as an option that nothing is selected to select everything before.
The algorithm, keeps the first block as the starting position and tries to optimize the selected blocks that follow.

Depending what you want to do you can first re-order the block you want to start with and then select what you want to optimize

Effer:
@vlachoudis I like the selectAll() if nothing is selected.

Maybe it could be useful to add also a way to invert selection (Ctrl+I ?) that I feel missing from the editor. Also add a couple of button for selection will make the commands more easy to find.

vlachoudis:
[...]

PR #199, opened 2016-01-16
Relaxed the constrains for block splitting when importing gcode

carlosgs:
Solves #179 
This is basically incorporating support for FlatCAM drill files, where lifting Z from a PCB substrate is performed with G1 instead of G0


carlosgs:
@vlachoudis  I'll make separate a pull request with the spanish translation only, sorry for mixing both commits

PR #200, opened 2016-01-17
Multipass

atrueresistance:
Multipass plugin to take selected code in the editor and copy with Z value transforms. This is very helpful if using bCNC for laser vector cutting on a lower powered system where more then one pass is needed to cut through material. 

1) Select code to copy
image

2) Fill out parameters
image

3) Editor will now have new blocks with correct Z depths
image

### DIFF
--- a/plugins/multipass.py
+++ b/plugins/multipass.py
@@ -60,10 +60,12 @@ class Tool(Plugin):
 		zreg = '(Zz? ?-?(\d+(\.\d+)?))'
 		zregexp = re.compile(zreg)
 
-
+		pos = blocks[-1]
+		pos += 1
 		for block in blocks:
 
 			newblock = []
+			newblocks = []
 			for passnum in range(0, int(np), 1):
 				for line in app.editor.gcode.blocks[block]:
 					z = zregexp.search(line)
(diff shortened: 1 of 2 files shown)


vlachoudis:
Thanks @atrueresistance. It seems to me that this is a duplication of the "Cut" tool.

atrueresistance:
@vlachoudis I do agree it is similar. However in my testing did not find it to work correctly. 
image
Is "Cut" supposed to tack on Z values if none originally given?
image

Given the output below, one can conclude no is the answer to that question. 
[mushroom-cut.zip](https://github.com/vlachoudis/bCNC/files/95905/mushroom-cut.zip)
In my mind it should take the first pass without Z values and run each additional pass with Z values decremented to user parameters.

Original file, direct from ObjectWorks program
[DirectOutputFromObjectWorks-mushroom.zip](https://github.com/vlachoudis/bCNC/files/95911/DirectOutputFromObjectWorks-mushroom.zip)

vlachoudis:
Yes it does. The second line "g1 z-0.125 f500" is setting the z value for the pass, even if no z is given. If it is necessary that the first pass is done with no z-value or with z-equal to the stock-z. I believe it is more logical to add the functionality in the "Cut" tool when the laser scanner is enabled rather than adding an additional tool.

Effer:
Even if I'd like to see a new plugin. I'm agree with @vlachoudis on integrate the option to start from stock surface in the CUT tool, maybe also for non laser machine.

atrueresistance:
@Effer @vlachoudis  Is there anything I need to do or can do? I'm not sure where exactly to start as `cut` is not a plugin.

*Nice to Have:* Having `Cut` place additional g-code in a new block would be helpful for clarity. 

@vlachoudis  Thank you for the explanation of `g1 z-0.125 f500`. Does the `cut` tool translate 3D gcode (variable Z throughout the file) decremented as well?

vlachoudis:
The "Cut" tool is calling the "cutPath()" method in the CNC.py.
It does multipass at various depths respecting also the tabs.
We can add an option there or on the calling routine to start from z=surface (for lasercutter)
For the moment is only doing 2D cuts.

vlachoudis:
In view of the correction of @Effer to the cut tool I believe I can close this pull request

PR #201, opened 2016-01-19
Spanish Translation

carlosgs:
Most of the translation is complete, the only missing lines are a few of the paragraphs with detailed GRBL alarm descriptions.

captura de pantalla de 2016-01-19 01-40-00


Effer:
Great work @carlosgs, bCNC is definitely going international.

PR #203, opened 2016-01-20
Proposal to open the user's configuration file.

Effer:
To avoid always the mismatch between bCNC.ini file and user .bCNC file when asking for the configuration file.
It's just a draft. I've tested only in windows with a machine with already associated .bCNC files.


vlachoudis:
Good idea!

vlachoudis:
The only worry I have is that if someone tries to edit the file from there it will have no effect since it will
be overridden by bCNC on exit.

Effer:
That true. What about open a simple tk window with a text field editor? In that way it could be possible to ask the user if want to save.
Or maybe better, integrate the whole thing like "shortcuts" or "machine".

Something must be done because every time user miss the right file.

PR #208, opened 2016-01-21
Cut from top?

Effer:
As discussed on #200 adding a way to cut from top as option.

image

or 

image

Still missing the grouping of added layer as request from @atrueresistance


vlachoudis:
I am wondering if we start from top, shouldn't the last pass be at depth-stepz?

Effer:
I don't think so. I prefer to consider the height of the material as target depth and stop. Top start is more like a test pass on non uniform material. That just my opinion

PR #218, opened 2016-01-30
Removed unused method and reference to Fluka.

Effer:
What's Fluka?


Effer:
Thank you to implements even update features!

PR #229, opened 2016-02-08
IT translation updated

Effer:


onekk:
Thanks, @Effer, I'm busy with some project and didn't have time to update the translation. 
I think I will be busy for some other time, but I'm not out of the business, only in pause.

Thanks to all and many thanks to @vlachoudis for creating bCNC.

PR #239, opened 2016-02-13
Buffer fill indicator

Effer:
Hi,
this is just a propose to have some sort of indicator of the grbl buffer state. Here it's the green pie graph on the bottom right:
image

I know is not really nice but with big files the green and orange bar quite overlaps and I'm often in doubt if I'm streaming too slow to the controller. 
Any improvement is welcome.

### DIFF
--- a/Sender.py
+++ b/Sender.py
@@ -128,6 +128,7 @@ class Sender:
 		self._pause      = False	# machine is on Hold
 		self._alarm      = True		# Display alarm message if true
 		self._msg        = None
+		self._sumcline   = 0
 
 	#----------------------------------------------------------------------
 	def quit(self, event=None):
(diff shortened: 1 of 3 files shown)


vlachoudis:
It could be useful, especially when not running, e.g. sending commands like jogging. However I don't like to incorporate it inside the ProgressBar. I use this widget on other projects not related to bCNC, and I would prefer to keep it simple.
Maybe it can be placed next to the progress bar as a separate widget.

Effer:
Ok, good suggestion. I'll try to create a new widget and update the pull request.

Effer:
Just moved to an external widget, first attempt with tkinter stuff. I'm not sure if the pack() is correct.
I'd like to change the color based on buffer level (green to red) but not sure which level should be considered critical (70% green, 50% yellow, 30% red ?)

image

Issue #153, opened 2015-11-18
French translation


ThierryM:
Hi,
Thanks for your work and sharing.
How could I add a french translation of the GUI or in an other language ? Of course, it's for sharing.
Regards,

Thierry

vlachoudis:
I didn't think of translating to other languages :) I saw that can be done using the gettext python module. I have to read how it works and let you know.

ThierryM:
Thanks for your reply. I will help with pleasure.

onekk:
Me too @vlachoudis, I volunteer for Italian Translation (I've done an Italian Translation and an Italian Guide for CamBam).

vlachoudis:
I've added a makefile that generates the .pot needed for the translation.
@ThierryM  I've made for testing the French .po and .mo files using the [poedit](https://poedit.net/) program
@onekk you can create a subdirectory it/LC_MESSAGES and do there the Italian translation.
it should go on the bCNC domain.

I've did quite quickly the change with the _("") infront of the strings that needs to be translated, so I wont be astonished if I missed something, or I've added it somewhere else which I shouldn't.

For the moment you need to set the environment variable LC_ALL to change the language
like
`export LC_ALL=fr_FR`
I will create an option inside the setup to change the language from bCNC directly, and independent from the OS.

ThierryM:
Hi,
Thanks for your work. I'm going to glance at the .po file.
Regards

onekk:
Hi @vlachoudis I'm translating bCNC but after cloning the dir I've modified (wrongly) the bCNC.pot file, but now I'm trying to regenerate it issuing "make pot" but it won0t work, what I'm doing wrong?

vlachoudis:
Do you have the xgettext installed?
Why not revert back the pot from the repository?

onekk:
Hi, Yes I have xgettext installed and even poedit i will try to revert it back maybe only the pot file, i will study the git documentation.

vlachoudis:
You can open directly from poedit the pot and then save it as  it/LC_MESSAGES/bCNC.po

onekk:
Ok I'm working on it

ThierryM:
Hi,
I have ended the french translation (it's not perfect but I think it works) with Poedit 1.8.6 on Ubuntu 14.04 (hard to compile it...). So how to integrate the french files bCNC.mo and bCNC.po in bCNC ?
[...]

PR #253, opened 2016-03-01
Update bCNC.po, see https://github.com/vlachoudis/bCNC/issues/153

tuxun:
The "È" char doesn't seems to be part of unicode.


vlachoudis:
thanks

ThierryM:
Hi,
As I told in the french translation discussion, I don't agree this commit because È existes in Unicode and in french, the capitalized characters have to be accentued.
So for me, this commit is a regression.

PR #264, opened 2016-03-06
Independent change of step and zstep in controlframe

HarryT36:
Hello,

first of all thank you for your great work.

I want to propose some changes in the behaviour of the controlframe for
jogging in x/y and z direction IN THE CASE that zstep is enabled in the
.bcnc file.

1. Independent INC/DEC/MUL/DIV Buttons for x/y and z
In the moment the values step and zstep are only individually changeable
by the combobox. I think it might be better if they are also
individually changeable by the buttons.

2. shift of the z-related buttons to the right side of the frame
So I can insert the new buttons on the most right side without changing
the rest of the frame too much. There are some "|" symbols to show when
x/y and z are seperated.

3. New button for move to a most up z-position
I use this often to clear y workspace on the CNC maschine. I use "G53
Z-2" which will work fine after "homing". For this button I also
inserted a new UNICODE symbol to unicode.py.

### DIFF
--- a/ControlPage.py
+++ b/ControlPage.py
@@ -375,28 +375,23 @@ class ControlFrame(CNCRibbon.PageLabelFrame):
 	def __init__(self, master, app):
 		CNCRibbon.PageLabelFrame.__init__(self, master, "Control", app)
 
-		row,col = 0,0
-		Label(self, text="Z").grid(row=row, column=col)
-
-		col += 3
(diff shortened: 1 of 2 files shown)


vlachoudis:
Thanks for the pull request. I will have to put it on hold for the moment to think about it. Normally I would like to reduce the amount of buttons currently present in the interface. So I am very skeptical adding buttons that the functionality can be achieved with what is already there.

PR #274, opened 2016-03-28
Fix Camera config menu

Effer:
Otherwise is not possible to config properly the new camera options.


vlachoudis:
Thanks @Effer for the moment is intentionally left disabled, since there is a conflict in some options of the config and the Probe camera

Effer:
Wow, it was something like an Easter egg..

szechyjs:
any status on re-enabling these config options?

PR #280, opened 2016-04-03
add german translation

feistus:
Not yet finished, but if you like, I would be glad to contribute to this great piece of software :)


Effer:
Hi @feistus , and finally a German translator arrived :)

vlachoudis:
Thanks!

PR #285, opened 2016-04-06
Camera in canvas prevent program to be closed

Effer:
At least in windows, if you let camera turned on in canvas, bcnc doesn't close nicely. The GUI close but the python prompt stay there forever.
@vlachoudis there could be better way to handle this. This is just for reference.


vlachoudis:
Thanks @Effer for checking on windows.

PR #287, opened 2016-04-08
Fix in sending overrided zero feed

Effer:
I've added the bug, now it should be fixed.
It happened when setting overrride before streaming anything, so it try to send a zero speed feed. That's no good.


vlachoudis:
I don't understand. How can it become zero? The scale only accepts from 25 to 200

Effer:
It's self._lastFeed that's is zero, if no feed has never been encounter before apply override. So new feed will be override*0 and is something GRBL doesn't like.
To reproduce: 
open bCNC
move override to 200%
open a file
start
...
grbl get angry

vlachoudis:
Ok, thank

Issue #291, opened 2016-04-11
Cant see the probe numbers (Change color)


sgtbidu:
How can I change the light green color of the probe numbers to a darker one? Its impossible to read them. Thanks in advance!

sgtbidu:
I searched for "green" in the CNCCanvas.py and replaced for "DarkGreen", and now is perfect.

PR #292, opened 2016-04-11
Master probe text color

Effer:
As suggest in #291 by @sgtbidu , added a user selectable color for probe text data

image


sgtbidu:
Thank you very much guys! Just today I manage to test this feature and works as a dream! Thanks a lot!

Effer:
You're wellcome. Any suggestion help bCNC grown better.

PR #295, opened 2016-04-12
Overrided by fit2screen

Effer:
canvas.reset seems not necessary


vlachoudis:
Unfortunately not. There is a draw command before, which is needed to calculate all extends of plot.
The draw command should be done with zoom=1.0 otherwise you might get crazy results.

Effer:
Got it. Thanks for explanation.

PR #305, opened 2016-04-20
Add custom color to block of gCode

Effer:
Allow custom color to block of gcode. Still missing the GUI side to adjust it.


Effer:
It would allow something like this for halftone with color:

image

vlachoudis:
Maybe we can use also the colors from dxf.

PR #328, opened 2016-05-21
Change block of colors

Effer:
I hope to have kept a coherent style with the rest of the code. Encountered some difficulties on understand the editor stuff.

I was thinking that the only context menu now present in bCNC is the one on the state machine button. Wouldn't be a nice GUI improvement to have all the editor Ribbon Menu show as a Context Menu when Right Click in the canvas?

### DIFF
--- a/CNC.py
+++ b/CNC.py
@@ -2722,6 +2722,14 @@ class GCode:
 		return undoinfo
 
 	#----------------------------------------------------------------------
+	# Set block color
+	#----------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


vlachoudis:
Thanks for the commit. 
Context menus was in my mind some time ago, but then I forgot :)

Effer:
Sorry for the horrible mess of commits.. I've steal some time from different place and pc, and I'm still a noob with git.

PR #338, opened 2016-05-27
camera-fix

Effer:
Without this, camera doesn't show in the canvas in my pc.

### DIFF
--- a/Camera.py
+++ b/Camera.py
@@ -50,6 +50,7 @@ class Camera:
 	def start(self):
 		if cv is None: return
 		self.camera = cv.VideoCapture(self.idx)
+		s, self.image = self.camera.read()
 		if self.camera is None: return
 		if not self.camera.isOpened():
 			self.stop()


vlachoudis:
Thanks. On linux it was working Ok, I don't understand why windows fails.
I've move it after the check on the camera for None.
However it should be better to be after the self.set()
and maybe a call on the "self.read()" instead to process the image as well

Effer:
To be honest I didn't dig too much. I remembered the camera was working some versions ago. Yesterday I made a small test and no picture was show in the canvas.
For sure the call should be done after the None test, but strange enough call to self.read() doesn't work. Neither after self.set(). No errors BTW, simply no images. I suspect the camera.isOpened() method have some differences in windows.

Off topic: that self.matchTemplate() looks interesting. Some time go I was testing an automatic zero reference based on a glyph placed on stock material.. First tests were not promising.

vlachoudis:
matchTemplate() I used it to calibrate the camera vs workspace, however mine has a big distortion so I didn't get nice results. First I need to calibrate the camera with a checker pattern and then I would be able to use the matchTempalte()

Effer:
I was testing with SIFT and FlannBasedMatcher. Quite slow and unstable. But a funny CV experiment.

Issue #345, opened 2016-06-02
DRO decimal place setting


tklus:
Is there a way to set the decimal place to 0.000 or 0.0000 all the time. for example, if the read out is displaying .5, I would like it to read 0.500 or 0.5000, if it is displaying .25 to read .250 or .2500. if it is reading 0 it would display 0.000

the DRO currently drops the trailing zeros.

It's not a huge deal but was curious if there was a setting for this.

Thanks,
Tim

Effer:
Hi @tklus ,
the option is missing right now. It can be easy added, I can give it a try next days. I like the suggestion.

tklus:
Nice work guys! The new setting works great! Should it be defaulted as 3 or 4? At the current 0, when you jog the machine it looks like it is not moving on the DRO because it is rounding to 1.

I would think most users would want either .000 or .0000 

Thanks again guys for the awesome support! I foresee this interface being very popular!

Effer:
@tklus just PR to fix that. Thanks to point it out

PR #349, opened 2016-06-04
DRO zero padding

Effer:
As point out by @tklus in #345 it wolud be nice to customize the padding of the DRO.  
Like here I believe: 

image

The settings will be set here (Tools->Machine->DRO Zero padding): 

image

Then restart bCNC to take effect.

### DIFF
--- a/CNC.py
+++ b/CNC.py
@@ -647,6 +647,7 @@ class CNC:
 	stdexpr        = False	# standard way of defining expressions with []
 	comment        = ""	# last parsed comment
 	developer      = False
+	drozeropad     = 0
 	vars           = {
 			"prbx"      : 0.0,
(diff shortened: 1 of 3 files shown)


vlachoudis:
Thanks. I was wondering we have already the 'decimal digits' option the rounding for the generated gcode. Shouldn't we use the same. I understand it is more flexible having it separated, but maybe it increases confusion

Effer:
That was in my  thoughts too. But I saw no better way to allow user to have the old DRO style or the new one. For me we could even keep a fixed formatting choice. It's seems like the all other DRO out there have a fixed digit look.

vlachoudis:
It is nice to explicitly set the variable also on the default bCNC.ini so if it is unchanged to avoid populating the users one

PR #377, opened 2016-07-16
4 terms translated in french

ThierryM:
Hi,
New french translation (only 5 new terms).
I couldn't translate "DRO zero padding" because I don't understand the meaning and the action provided.
Regards,

Thierry


Effer:
Hi @ThierryM ,
DRO zero padding is the number of fixed zero added to the left in the DRO. Some users find it easy to read it with explicit zeros.

This is with DRO zero padding = 0
image

and this with DRO zero padding = 3
image

Cheers

ThierryM:
Hi @Effer ,
Thanks for your explanation. In fact, it's just a display option (number of digits) but the given name is too complicated according to me. Sometimes, simpler terms are easier to understand : DRO for Digital Read Out isn't explicit (almost for a beginner like me).
For me in french, I would translate in "Nombre de décimales sur l'afficheur".
Regards,

Thierry

PR #379, opened 2016-07-19
Input Wpos from Float Entry

Effer:
Dirty attempt to allow user input some calculation in the WPOS input fields.

Looking at the code I've notice that the **setY** method in ControlPage.py was wrong (maybe some test left behind?)

Also, during the input the **HOME** key is not disabled. Probably  this should be avoid or it'll start the Home procedure instead of changing the caret position.

image


vlachoudis:
Thanks

PR #395, opened 2016-08-04
Fix again InPlaceText in windows

Effer:
In windows the self.toplevel.update_idletasks()  is needed. Otherwise the edit field is not visibile, even if it records any keystroke.
It was already committed here https://github.com/vlachoudis/bCNC/commit/9c0c5020e5ce6d53e6f3ee5dda48c4981a6b5723

Is there any problem with this in Linux?


vlachoudis:
I generally avoid to use the update_idletasks. I will check more carefully before adding it.

Issue #396, opened 2016-08-05
Report at job end


DasWookie:
One of the neat things about Universal gCode Sender is that at job end, it reports how long the run took. There was always some other information I felt it could also have contained such as lines executed, start and finish time, complete status  such as did it finish 100%, aborted because of hard stop, canceled at line number XXX, et cetera. I could then use this information for comparison against the run time estimate, as well as see when the job completed so on subsequent runs I'd have some idea of when I needed to poke back in, swap the stock, and start the next job.

I figure I'm probably not the only person who would like to (possibly) see this sort of end of run report, so thought I'd throw this enhancement request out there. Maybe get fancy and have it be possible to disable it for those who don't want to see it (if it's a modal dialog box) or just toss it into the console output (or status line on the bottom left).

DasWookie:
That end of run report change with the Events change nailed just what I was after. Thanks! :)

PR #397, opened 2016-08-06
Master vasilis run end info

Effer:
- Verbose output as suggested in #396 
- Pending PR #395

### DIFF
--- a/Sender.py
+++ b/Sender.py
@@ -13,6 +13,7 @@ import re
 import sys
 import rexx
 import time
+from datetime import datetime
 import threading
 import webbrowser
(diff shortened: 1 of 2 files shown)


Effer:
Removed not related patch from this PR

vlachoudis:
I didn't succeed to merge with git, so I made it manually :) Thanks

vlachoudis:
The "runEnded" function is called when all lines are streamed, but it could well be that some are still pending to be executed in grbl. The same holds on the "onstop" event I just added. I need to wait until the status is Idle before executing them.

PR #399, opened 2016-08-06
Fix parseLine2

### LINKED ISSUE #398: class CNC has no attribute 'parseLine2' when manually sending commands
Sending a gcode file seems to work fine but when entering anything into the "Command" box, I get this error popping up.

error

Running on windows 10
>Python 2.7.10 (default, May 23 2015, 09:40:32) [MSC v.1500 32 bit (Intel)] on win32

Other python programs work fine, eg Printrun (pronterface)

Effer:
Fix #398 
Refactor went wrong?

### DIFF
--- a/Sender.py
+++ b/Sender.py
@@ -187,7 +187,7 @@ class Sender:
 	# can return a python exception, needs to be catched
 	#----------------------------------------------------------------------
 	def evaluate(self, line):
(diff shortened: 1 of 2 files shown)


Effer:
Get it wrong.

vlachoudis:
Indeed. However should be converted to compileLine and not to parseLine.
I did the correction

PR #413, opened 2016-08-30
Master vasilis auto load

Effer:
I'm always in doubt if the file has been automatically reloaded by bCNC when modified externally.
This is a propose to make it more verbose, please fix style and any function sign if you don't like it.
(and let me know for the next time :)

image


vlachoudis:
Thanks. That good indeed.

PR #420, opened 2016-09-08
Automatically clear connected message.

jeffmd:
Instead of having the user clear the connected message it is now done automatically when a valid operation is sent to grbl, ie jogging the machine, selecting run/start.
This also corrects problems with run/idle not displaying when connected is displayed.  If the user doesn't clear the connected status message, or select stop or hard reset, then it always stays displayed.  
This also fixes a bug with pause/resume when connected message is displayed - pause would work the first time selected but selecting pause button to do a resume would not work.  Very annoying when using the pendant feature on a tablet since you have to type in resume as a command the first time.

### DIFF
--- a/Sender.py
+++ b/Sender.py
@@ -841,6 +841,9 @@ class Sender:
 						status = False
 						if not self._alarm:
 							CNC.vars["state"] = pat.group(1)
+						elif  self._gcount > 2 and CNC.vars["state"] == CONNECTED:
+							# turn off alarm for connected status once a valid gcode event occurs
+							self._alarm = False
 						CNC.vars["mx"] = float(pat.group(2))
 						CNC.vars["my"] = float(pat.group(3))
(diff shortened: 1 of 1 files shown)


vlachoudis:
Thanks for update. I want first to test on my machine (next week) before committing to be sure that it doesn't clear the message in case of true errors.

jeffmd:
Sounds good.

I checked the error/alarm code section in Sender.py and used grbl simulator to inject alarm/error states.   An alarm/error will always change CNC.vars["state"] to something other than CONNECTED and the change I made will only clear the _alarm if state is CONNECTED.  But, always good to get another set of eyeballs to check it out and test.

vlachoudis:
@jeffmd thanks for the pull request. I will have to check it on my machine (sometime next week) if it doesn't break anything else before applying it.

PR #421, opened 2016-09-08
Avoid accessing out of list index

Effer:
It may happen when canvas is not sync with editor gcode.


vlachoudis:
The existing try: except: should catch this condition

Effer:
I am quite sure that I get an exception on that line reported on the console.

PR #426, opened 2016-09-10
Making space for a 4th axis from grbl

gcobos:
Adds controls in the UI for an extra axis "E" and some logic to process messages from GRBL using either 3 or 4 axis.

This is a personal modification I did as an experiment, to be able to use bCNC for 3D printing.

The only thing that breaks what you have now, is that I use 'Home' and 'End' keys to control the 4th axis.


vlachoudis:
Thank you @gcobos. It would be nice to have an option in the settings to enable a mode for 3d printing as for the lasercutter.

Harvie:
I think it would make sense to start with Marlin support #1020 #455 
Bacause as far as i know you can't run 3D printer on GRBL

gcobos:
@Harvie This pull request is obsolete and it should be replaced by a new one soon. I have been able to 3D print with bCNC, but obviously with many limitations. That said, I think enabling bCNC to handle more axis makes it more flexible, allowing to work with more types of machines.

Harvie:
> This pull request is obsolete

Ok, i will close it then. For now. That doesn't mean i am not interrested in this topic.
What kind of controller did you used? GRBL does not support 4th axis as far as i know.

PR #427, opened 2016-09-11
modified some text in Italian translation

onekk:
Modified some test on probe page, the listbox is too small to visualize a meaningful test.

To partially bypass the behaviour of the listbox items to not display the full text in a tooltip if the text exceed the space.


Effer:
Hi @onekk , keep in mind that the GUI is resizable.
The user can drag the right border of the left panel, in this way it enlarge also the combo box. So maybe some text can be verbose.

onekk:
Yes, but without any intervention I have some text that is over my 1024x768 screen.

In the Tools "window" I have some icons out of the window.

How to resize a panel? 
How to customize the interface?
Maybe through some variables in ini files?

The main problem with bCNC actually is the lack of documentation.

Effer:
To resize panel: hang the mouse over the border (red in picture):

image

Cursor should change to a double arrow shape (not show in picture). Now Click and drag the border.

You can even collapse some panel if you click over the name:

image

All this is well **undocumented**.  Wiki is always late with development :)

onekk:
Filippo @Effer could you put the above information in a page describing the interface, I could not insert images in the wiki, It will be a better improvement to have a manual, for me it's the "most wanted" feature of a program and could spread bCNC usage.

Effer:
Ciao Carlo @onekk ,
here the page https://github.com/vlachoudis/bCNC/wiki/Customize%20GUI.
For the wiki picture there are two way to upload image. One is to post them somewhere where you can (for example a pull request) and the other is to add them in the doc branch https://github.com/vlachoudis/bCNC/tree/doc/Screenshots

The second, I believe, is the preferred way but, as any other PR, it must be approved. I found it more uncomfortable as it take more time. And lot of time is needed to full document bCNC.

onekk:
Many thanks, when I found the time, i will make a framework in Latex for a proper manual.

Regards

Issue #433, opened 2016-09-19
Cannot access C:\ folder


acaballero6270:
When trying to open a file that is located in folder C:\home. When I click "C" in the open file dialog it jumps to the root folder of bCNC instead of C:\ . Is that a bug? or am I doing something wrong?
Running on Windows 7 with python 2.7.11.

Effer:
Thanks @acaballero6270 to point it out.
I've found how to fix in windows but can't test in linux now.
If you want to modify the code by yourself have a look [here](https://github.com/vlachoudis/bCNC/pull/434/files).

acaballero6270:
It worked. Congratulations!

PR #434, opened 2016-09-20
Fix for #433 at least in windows

Effer:
If anybody could test in linux.
It's a sort of trick, but does its job.


jeffmd:
Yup, seems to work fine in linux, changing paths still works properly.

PR #441, opened 2016-09-24
Added preset steps s1 s2 and s3 and spindle dir and coolant/vacuum on/off

onekk:
Maybe it is rough, and i didn't know why this line is present in the original code

`		self.spindle.set(CNC.vars["spindle"]=="M3")
`
but it is not used in setting the spindle direction in **spindleControll** function

Plus the line 
```
 	def setStep3(self, event=None):		 	
 		if event is not None and not self.acceptKey(): return		 
		self.setStep(self.step3, self.step2)
```
 that sets the zstep = step2 in the step3 routine seem strange.

Maybe, and I'm guessing the differences between the Z axis is the reason to use step2 instead of step3?

Regards


vlachoudis:
Thanks @onekk I've added your steplist settings, but not the extra buttons for step1,2,3
Also the ones for the spindle direction and coolant I want somehow to merge them with the existing ones
to reduce the number of buttons

PR #447, opened 2016-10-01
auto retrieve min e maz from steplist

onekk:
Hope it will be useful.

Note it is ready for the separate zstep increase and decrease operations


Harvie:
Is this still a thing? Recently there were some changes in how stepZ is handled.

PR #449, opened 2016-10-02
Trying to get an infrastructure for multiple configuration files

onekk:
This is a first try to get an infrastructure for using multiple configuration files.

It define a global variable **glb_data** that holds a dictionary containing the "sections" of the configuration file, that could be parsed at the end to rewrite the configuration file in the appropriate place.

the dictionary holds a dictionary with the sections containing in the value a sub-dictionary holding the name,value pairs

It search in multiple path defined in the search_paths, that can be modified to use multiple OS dependant paths

the order is important the first file found is used.

the dictionary is easy adressable with `Utils.glb_data["Section"][name]` to obtain the appropriate value.

It is rough but it could be useful as a start, in this manner each plugin could have is own ini file with the pattern plugin_name.ini maybe in  `~/.config/bCNC/` on Linux and %PROGRAMDATA%\bCNC on windows (if i didn't go wrong with the windows location)

Regards

### DIFF
--- a/ControlPage.py
+++ b/ControlPage.py
@@ -26,10 +26,27 @@ import CNCRibbon
 from Sender import ERROR_CODES
 from CNC import WCS, DISTANCE_MODE, FEED_MODE, UNITS, PLANE
 
-_LOWSTEP   = 0.0001
-_HIGHSTEP  = 1000.0
-_HIGHZSTEP = 10.0
 
+try:
+	_LOWSTEP = min(map(float, Utils.config.get("Control","steplist").split()))
+except:
+	_LOWSTEP   = 0.0001
+
+try:
+	_HIGHSTEP = max(map(float, Utils.config.get("Control","steplist").split()))
+except:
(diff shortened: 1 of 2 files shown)


onekk:
I've found this library that could help to use different appdir on different platforms

[appdir library](https://pypi.python.org/pypi/appdirs)

vlachoudis:
@onekk your suggestion is nice. I would prefer not to use any additional library for that. I am working on a branch that it creates a subdirectory .bCNC instead of a file, and everything should be split there.
One option is to split into multiple files the configuration for every section and when it loads them to load everything in the same structure as I have now. This will minimize the code changes. Another option maybe better is to have several configparsers objects for each database.

onekk:
@vlachoudis, bCNC is a very good program and I reccomend it when some user has problem with gcode sender on other forums (CamBam forum mostly).

Every solution to the problem is welcome, how it's only a matter of taste and maybe some clear explanation and example files are welcome.

Thanks for your work and Best Regards

PR #472, opened 2016-12-26
Bugfix: pause button would not resume if job paused (grbl 1.1)

jeffmd:
- The bug only occured with grbl 1.1 since it now reports **Hold:0** or **Hold:1** instead of **Hold** as in older grbl versions. The fix does not effect operation with grbl 0.9.


vlachoudis:
Thanks

PR #480, opened 2017-01-03
Added Zig-Zag electrode plugin

CarlosGS:
Happy new year to all, and thanks for so many great improvements of bCNC :-)

Here is a plugin that generates a Zig-Zag pattern, it is meant for custom PCB finger electrodes. It differs from the "flatten" plugin in that the inner/outer corners can be rounded; with Zig-Zag it is easier to create structures that have a constant width gap and no sharp corners.
It also has a different set of parameters: rather than defining an area, the user can select the number of finger rows and their length.

image

image

Zig-Zag is nothing too fancy, but it might come in handy for somebody building DIY moisture detectors, chemical sensors, tactile buttons, or other similar wavy structures.

image

In this case I've built an electrode array that can detect ionizing radiation (google "spark particle detector). High static voltage is applied to the electrodes, and sparks form whenever an alpha particle interacts and ionizes air near the copper.
Here is a video showing a few spikes of background radiation: https://www.youtube.com/watch?v=s3N2EDGwe9U (only a proof of concept).
I know there are a few physicist in the room... please share your thoughts on the matter :)


vlachoudis:
Very interesting! How high was the voltage you applied?

PR #506, opened 2017-01-28
Double size icon in user settings

Effer:
Allow user to get monstrous icons on their High Resolution Display.

image


vlachoudis:
Thanks

Issue #419, opened 2016-09-07
Auto leveling on smoothieboard not working


modulab:
Hi there!
1st thing, thank you for making this software available. 
2nd thing, i've been having some trouble with the auto leveling sequence. Actually it never worked. After the probe's first contact i get an error: "Probe triggered before move" .Sometimes i get to probe two locations on the grid but that's it. Same error and full stop. 
I'm running on a smoothieboard, up to date with firmware and bcnc on windows 10. I have tried everything electrically : cap, shielded wire, em choke. 
I suspect it has to do with the difference between native smoothieware auto leveling or bcnc.

Can you please offer me some advice? It bugs me for so long now. 
Thanks, 
Paul

Effer:
Hi @modulab ,
is that exactly the error you read in bCNC? Do you read that message in the terminal output?
I can't test with smoothie but I can have a look to understand if is something due to bCNC or simply some electric noise.

pda3k:
I echo the thanks for this great piece of software.  It is my "goto" CNC controller software for the Smoothieboard.  I also experienced the exact same issue/error with the exact same conditions.  It has been a while since I looked so I don't remember if the message was in the terminal or not.  I seem to remember it being in bCNC.

Effer:
Can you post the output from the terminal tab when the probe routine fail?

![image](https://cloud.githubusercontent.com/assets/1007007/18760706/786f8838-8102-11e6-8192-ab32fa7db54c.png)

pda3k:
Below is the output as requested.  I am running version 0.9.6

G38.2Z-25F5
[PRB:123.262,-191.878,-10.310:1]
ok
G0Z3
ok
G0X31.739Y-32.879
ok
G38.2Z-25F5
error:ZProbe triggered before move, aborting command.
Run ended
2016-09-25 14:21:06.963000
Current: 16 [17]  Completed: 23% [7s Tot: 7s Rem: 0s]

ok

ok

ok

error:ZProbe triggered before move, aborting command.

ok

ok

ok

error:ZProbe triggered before move, aborting command.

ok

ok

ok

Effer:
@pda3k  just for test, can you post the output using this settings?

![image](https://cloud.githubusercontent.com/assets/1007007/18817502/42d567ee-8362-11e6-8304-668b023a6ef9.png)

[...]

PR #576, opened 2017-05-07
Fixed autoleveling for smoothieware (both freezes and G0 speeds)

MoonCactus:
You may be interested by this fix, which is aimed at a better cooperation with Smoothieware.
I think it should not impact other firmware.

The mesh leveling froze on the first probe (already in contact), but it works with an additional %wait.

Also, smoothie use the last feed ratio for its G0 moves (which is wrong). This makes the leveling extremely slow after the first probe (because probing redefines the F to a very low value). So I just added an explicit single-line F command to reset the "feed" value to the initial value before calling G0. I think it should have no effect on firmwares that handle G0 properly (at maximum machine speed, whatever current feed rate).

### DIFF
--- a/CNC.py
+++ b/CNC.py
@@ -275,17 +275,22 @@ class Probe:
 		self.makeMatrix()
 		x = self.xmin
 		xstep = self._xstep
-		lines = ["G0Z%.4f"%(CNC.vars["safe"]),
+		lines = ["F%g"%(CNC.vars["feed"]),
+			 "G0Z%.4f"%(CNC.vars["safe"]),
 			 "G0X%.4fY%.4f"%(self.xmin, self.ymin)]
 		for j in range(self.yn):
 			y = self.ymin + self._ystep*j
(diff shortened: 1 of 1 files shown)


MoonCactus:
BTW this is most probably related to #419

vlachoudis:
Thank you. The feed is ok, the additional waits, will simply impose a short wait until grbl reports idle, before starting the probe. I am merging the request, but I will check if I could add them only for smoothie

MoonCactus:
Thanks! Given that probing a slow operation I omitted the check for SMOOTHIE (that would pollute a bit the source code imho). A comment may be welcome anyhow, for other developers that could wonder what is going on, and may try to optimize it out!
And in any case thanks go to you! I really consider bCNC to be a good, clean and compact program. This already has become my prominent tool, especially as I know I will be able to add and tweak stuff easily.

vlachoudis:
You are welcome.

PR #579, opened 2017-05-09
Fix bug where bCNC is sending f0.4$X

eyal0:
This fixes a bug for me though I don't complete understand why.  The procedure to reproduce the bug for me:

1. Open bCNC, connect to router.
2. home, reset, unlock
3. Calibrate.  Let it complete or cancel mid way, doesn't matter.
4. Alternate between clicking on reset and unlock buttons.

The error "expected command letter" appears when unlock is pressed after reset.  If unlock is pressed again, then it doesn't happen.

It seems that bCNC is sending "f0.01$x" to the router.  I also see it if I try to tool change, in which case bCNC sends "f0.01$g".  This PR fixes the bug for me.

Let me know if you need more info.

### DIFF
--- a/Sender.py
+++ b/Sender.py
@@ -964,7 +964,7 @@ class Sender:
 							CNC.vars["_OvChanged"] = False
 							self._newFeed = float(self._lastFeed)*CNC.vars["_OvFeed"]/100.0
 							if pat is None and self._newFeed!=0:
-								tosend = "f%g" % (self._newFeed) + tosend
+								tosend = "f%g\n" % (self._newFeed) + tosend
 
 						#Apply override Feed
 						if CNC.vars["_OvFeed"] != 100 and self._newFeed!=0:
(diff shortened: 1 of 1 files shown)


vlachoudis:
Appending a new line will break the "ok" counting and result in miss match, in the submitted lines and what bCNC thinks was processed from grbl. it has to be done in another way.

eyal0:
Can you suggest a better way?  I don't understand the code well enough to make the change.  Maybe in the function `softReset` we can see if there is a new feed speed and, if so, `send.sendGCode` with the new speed?

vlachoudis:
I was thinking maybe with a condition, to add the F## only only if the tosend command doesn't start with '$'.
To modify the line above to something like 
```
if pat is None and self._newFeed!=0 and tosend[0]!='$':
```
However I cannot test it right now.

eyal0:
Okay, I did it that way and the fix is working for me.  Take a look?

vlachoudis:
Ok, I've copied manually since it was faster than pulling

PR #596, opened 2017-05-24
Add fastfeed option for faster calibration and tool changes.

eyal0:
Currently, the probe feed rate is used for tool changes, calibration, and autolevelling.  When autolevelling, it is useful to go slowly so that the tool doesn't damage the PCB surface.  But for tool change and calibration, the sensor is spring loaded and it's possible to move the tool toward the sensor much faster.

However, moving the tool too fast results in inaccurate results.  So instead, start with a fast probe, then back off and try again 10 times slower, over and over, until doing the slowest setting.  This allows for faster probing without sacrificing accuracy.

If the fast feed is unset or is less than or equal to the regular feed speed then this option will have no effect on operation.

### DIFF
--- a/CNC.py
+++ b/CNC.py
@@ -1671,6 +1671,15 @@ class CNC:
 			lines.append("g53 g0 z[toolprobez]")
 
 			# fixed WCS
+			if CNC.vars["fastprbfeed"]:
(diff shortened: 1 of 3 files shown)


vlachoudis:
Thanks

PR #609, opened 2017-06-26
Fixed transformation order for camera rotation

dguerizec:
Rotation first, then translation.
    This is much easier to set up.

### DIFF
--- a/Camera.py
+++ b/Camera.py
@@ -109,19 +109,20 @@ class Camera:
 	def rotate90(self, image):
 		if self.rotation > 0:
 			rows, cols = image.shape[:2]
+			m = cv.getRotationMatrix2D((cols/2, rows/2), self.rotation,1)
+			image = cv.warpAffine(image, m, (cols,rows),
+					None,
+					cv.INTER_LINEAR, cv.BORDER_CONSTANT,
+					(255, 255, 255))
+			rows, cols = image.shape[:2]
 			t = np.float32([
 				[1, 0, -self.xcenter],
 				[0, 1, -self.ycenter]])
 			image = cv.warpAffine(image,t,(cols,rows), None,
-				cv.INTER_LINEAR, cv.BORDER_CONSTANT,
-				(255, 255, 255))
-			rows, cols = image.shape[:2]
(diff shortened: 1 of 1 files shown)


dguerizec:
Short documentation, since I just had the opportunity to test it in real situation:

1. Jog the gantry to align the camera on top of a "vertical" or "horizontal" line (along the X or Y axis).
2. Set the rotation in the Probe/Camera form, so that the line is straight.
3. Jog the camera on top of a point.
4. Jog the gantry as low as possible, and adjust it so that the point is in the center of the haircross.
5. Jog the gantry as high as possible.
6. Adjust the Haircross Offset values (X and Y) so that the haircross match the point.
7. Go back to 4., repeat until the haircross match the point at any Z altitude.

Alternate method to find Haircross Offset:
3. Place a mirror flat under the camera
4. Jog the camera as high as possible
5. Adjust Haircross Offset to match the haircross and the camera lens

vlachoudis:
To match to any height can be done only if the camera is parallel with the Z axis, which can be quite a challenge to do.

vlachoudis:
Could you please add it to the wiki.

dguerizec:
Yes, there is a point that is on a perpendicular line if the camera is not too tilted.
The remaining problems are the asymmetric distortion, and the scale regarding to the Z height.
OpenCV should be able to handle that, I'll check it when I have some time.

And I'll add it to the wiki.

vlachoudis:
One confusion now, is that there are still both parameters the old 90deg interval angle orientation and the new rotation. Maybe the old one can be removed since it is redundant. One minor issue is the dimensions of the rotated image keep the initial dimensions so they are cropped. The effect is more impotysny if you rotate by 90deg.

dguerizec:
Yes, there are both parameters for now, and the "advanced calibration"
options are disabled when rotation is 0.

This is still experimental, so I want to encourage people to test it, but
I'd also like to let people use the old method until parameters are well
defined.
I'll remove the old angle parameter when I'm done.

For the cropped image, I plan to work on this too, but it is not a big
problem for the moment.

2017-06-27 11:47 GMT+02:00 Vasilis Vlachoudis <notifications@github.com>:

PR #655, opened 2017-09-07
Fix broken "Garbage Receive" statements missing a %

m-g-m:
Any "Garbage" data was causing bcnc to throw an error and go into a zombie state. The "Garbage Received" string formatting clauses need a % between the strings and the values.


vlachoudis:
Thanks!

PR #662, opened 2017-09-25
Add functionality allowing one to specify camera settings via one's INI settings.

coddingtonbear:
This adds functionality allowing users to customize the capture settings of their attached webcam by setting specific INI properties.

I added this functionality because the camera I use for CNC alignment captures at a high-enough resolution by default that bCNC became unnecessarily slow.  What this feature provided for me was the ability to override the default capture resolution to set it at a more reasonable value.

Note that I've tested only the `height` and `width` properties specifically, and not all cameras support all provided properties.  The other properties are provided because the settings were in nearby documentation and adding the extra options was minimal effort. 

Cheers!

### DIFF
--- a/Camera.py
+++ b/Camera.py
@@ -41,10 +41,56 @@ class Camera:
 		if cv is None: return
 		self.prefix  = prefix
 		self.idx     = Utils.getInt("Camera", prefix)
+		self.props   = self._getCameraProperties(prefix)
 		self.camera  = None
 		self.image   = None
 		self.frozen  = None
 		self.imagetk = None
+
+	def _getCameraProperties(self, prefix):
+		""" Gather user-defined camera configuration properties
+
+		Allows user to specify parameters like::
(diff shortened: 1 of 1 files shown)


vlachoudis:
Thank you! Could you please document it also in the wiki page.

vlachoudis:
Apparently in the new cv api the variables should not use the CV_ prefix and peaked directly from the cv2

coddingtonbear:
Ahh -- you're right -- I _do_ see that in some of the docs; I'll try to find a little time tonight to have a look at what it would take to make this properly handle both the old and the new CV apis.

coddingtonbear:
Oh, nevermind, I see that you made the changes for migrating to the 3.x API names!  It looks like I'm just using an older version of OpenCV where I have this installed; I'll just upgrade mine.

Thanks for fixing the problem!

PR #669, opened 2017-10-07
Fix Z when backing off the probe

chaotix-:
Pulling back [tooldistance] in relative mode may end at a coordinate
outside the working area. At least GRBL 1.1 will throw an alarm and there
is no way to reset it except closing the serial port and reconnecting (and
loosing microstep positioning by doing this).

This changes pulling back to the absolute Z position we know is good: [toolprobez].


chaotix-:
Sorry. This is not ready yet. Now I get an ALARM 4, error 9 one line after backing off. Will investigate further.

PR #672, opened 2017-10-08
driller laser mode compatible

Effer:
Better compatibility with laser mode


vlachoudis:
Thanks @Effer

Effer:
You're always welcome @vlachoudis

PR #677, opened 2017-10-12
Adds functionality showing the displayed CNCCanvas in Pendant.

coddingtonbear:
Hey there @vlachoudis -- I hacked together this little bit of functionality just to scratch my own itch, and am submitting this PR just in case you might find such functionality useful.

I'm not totally sure I'm on-board with merging this as-is -- it's a little bit of a hack, honestly, but I did briefly consider adding configuration options allowing this (and a couple other existing features) to be configured from the `ini` file.  In the end, I decided to table the idea of adding the configuration options because I thought it might be better to consider swapping out some parts of the hand-brewed web server with something providing a little more out-of-the-box functionality.  I'd be glad to chat with you about what I can help you with if you were on-board with such an idea.

Cheers!

*Example*:

image

### DIFF
--- a/Pendant.py
+++ b/Pendant.py
@@ -13,6 +13,7 @@ import sys
 #import cgi
 import json
 import urllib
+import tempfile
 import threading
 
 from CNC import CNC
@@ -21,6 +22,11 @@ try:
 	import urlparse
 except ImportError:
 	import urllib.parse as urlparse
+
+try:
+	from PIL import Image
+except ImportError:
+	Image = None
 
 try:
 	import BaseHTTPServer as HTTPServer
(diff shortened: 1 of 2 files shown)


vlachoudis:
Thank you I liked that! I am wondering if there is a more sufficient way to pass from the canvas to bitmap rather via a postscript file?

coddingtonbear:
Probably!  I didn't look into it in much depth because I wanted to check-in with you first before digging too far into possibilities -- there are lots of things I'd love to work with you to polish-up.  I wonder if you'd have some time to chat on #freenode sometime about some ideas I have for polishing-up pendant?

PR #685, opened 2017-10-23
Adds rudimentary automated Travis CI tests.

coddingtonbear:
Hey there @vlachoudis --

I was considering working on making bCNC compatible with Python 3, but couldn't really get started without having a way of verifying that I hadn't broken anything.  Given that bCNC doesn't appear to have much in the way of tests, I've put together a limited test suite providing a couple "smoke" tests.  It doesn't cover much, but it's at least a start.

Things automatically tested by this suite in Travis CI:

* Launching the GUI.
* Connecting with a GRBL simulator (via [grbl-sim](https://github.com/grbl/grbl-sim)).
* Loading and running some sample Gcode.

If you aren't familiar with Travis CI, it is a very commonly-used free continuous integration service allowing you to automatically run tests as pull requests are posted or changes are pushed.  After merging and enabling the integration, it will automatically kick-off a build of the test suite -- here's what that looked like when I was working on putting these tests together: https://travis-ci.org/coddingtonbear/bCNC/builds .

I've also set this up to generate animated gifs of each test run to aid in troubleshooting test failures if they were to occur -- here's an example of one:
here is an example of one

If you do merge this, there're just a couple one-time setup  steps to perform:
* Using your github account, log-in to https://travis-ci.org/auth
* Go to your profile page and enable builds for vlachoudis/bCNC
* Add a limited access AWS access key, secret, and bucket to this repository's Travis CI configuration by following these instructions: https://docs.travis-ci.com/user/uploading-artifacts/ -- if you do not have an AWS account or would rather not use yours, I'd be glad to provide you with access to a bucket on my AWS account for this purpose.

Cheers, and feel free to reach out if you have any questions!


vlachoudis:
Thank you. I will certainly give it a look. I don't know how travis works, so it will be a great opportunity for me. Thanks

coddingtonbear:
No problem -- I think it'll help a ton for anybody working on adding new features by making it very easy to verify that nothing might've been accidentally broken as part of a feature's development.  

Let me know if you run into any questions that you think I might be able to answer for you -- I'm 'coddingtonbear' on Freenode, and me@adamcoddington.net on google chat if you'd like to have a conversation about anything.

Cheers!

.gitignore
```diff
@@ -1,3 +1,13 @@
 *.pyc
 lib/*.pyc
 .idea
+tests/bin/*
+tests/lib/*
+tests/include/*
+tests/.Python
+.cache
+bin/*
+lib/*
```
Harvie: I am willing to merge this. But i don't like idea of puting lib/* in .gitignore as we use this directory and it will lead to confusion. do you really need to ignore everything in that directory?

coddingtonbear: Ahh; I added that directory because that's where the virtualenv stores installed python dependencies.  Let me have a look at this again over the next few days to see if I can update this to move that into a different spot.  This PR is also a little old now and I'll try to make some time to update it to the current standards.

Harvie: ok. thanks!

coddingtonbear:
I'm having just a ton of trouble getting the grbl simulator to behave reliably; so I'm afraid I'm at a little bit of an impasse at the moment.

Currently -- and I vaguely remember occasionally running into problems with this when I wrote this PR last year -- the simulator will occasionally hang at some point along the process of running a gcode command.  At the moment, I'm chalking this up to some sort of subtle timing difference between the grbl simulator and grbl as run on actual hardware, but I'm honestly not very confident about that.

There're at least a few ideas I have offhand for moving forward:

* Abandon the idea of using `grbl-sim` and instead write a simple (probably python) simple gcode listening script.
    * This is probably the avenue with the highest chance of success, but provides the weakest guarantees in that the running test wouldn't actually be evidence of actually being able to interact with a grbl device  (that being said, right now the genuine `grbl-sim` doesn't appear to offer that kind of guarantee anyway).
    * If this idea were to be pursued, I'd probably consider it a short-term solution to allow us to at _least_ test the GUI, and continue investigating more-accurate simulations.
* Investigate using `grbl-sim` with one of the less resource-constrained avr versions.  I'm not totally sure if `grbl-sim` works with the atmega2560 grbl variant or not, but that might be a possible path forward. 
* Investigate using `simavr` instead of `grbl-sim`.
    * I actually did try this, and ran into similar (maybe the same?) problem as I had encountered when using `grbl-sim`.  It's worth another look, though, of course; I could have easily botched something in my configuration.
    * Like the above option, it might be useful to try running one of the less-resource-constrained grbl implementations on `simavr`.  I'm fairly sure `simavr` can also simulate an atmega2560.

I'm for the moment setting this on my back burner given that I've spent quite a lot more time than I'd expected to already on this, but I'm curious about what ideas or thoughts you might have about the above options.

Harvie:
> Abandon the idea of using grbl-sim and instead write a simple (probably python) simple gcode listening script. 

I was hoping to get this for quite some time. Currently i always have arduino with grbl (but no motors or connections) on my table to be able to debug bCNC. That kinda sucks, because i don't have any buttons to simulate probe, so i have to short the input pins to ground using scissors. That's far from ideal.

Currently i've managed to separate 99% of controller specific code to separate plugin files #1020. One of the resons why i did this was to enable us to do this exact fake GRBL thing.

But i've just realized this does not belong to GRBL handling plugin. GRBL plugin should not be replaced, but it should rather be connected to fake GRBL (so we can test if the GRBL plugin works).

So i propose that we separate serial connection to separate plugin, so we can use different (or fake) means of connection to GRBL. There are already reasons for this:

  - one connection plugin will use serial as we know it
  - another one will connect to serial over IP/telnet (as supported by GRBL ESP32 clone)
  - another one will not connect anywhere, but rather fake the GRBL responses in some basic manner (i guess that we just need to update WPOS/MPOS using coordinates received from bCNC, in the beginning we might even ignore the arcs and do straight line to that coords. although G90/G91 and WCS support might make some sense)

Harvie:
> I was considering working on making bCNC compatible with Python 3, but couldn't really get started without having a way of verifying that I hadn't broken anything.

Do you have any experiences with porting python2 to 3 ? I've been trying some automatized scripts, but i was unable to fix the resulting code. Few days ago i've updated `vlachoudis:v3` branch to reflect latest development of bCNC, but i only got it to show the main window, but most features are broken. There are lots of problems and lots of them seems to be related with Tkinter and how it's internal objects and string representations work with new unicode strings. Automatic migration scripts were not able to fix this and i was completely baffled by it.

I have created milestone for python3 https://github.com/vlachoudis/bCNC/milestones
First i want to have one stable python2.7 release and then we can start working on python3.

Also i would not be worried about testing everything perfectly. I've done some major refactoring in this month and what i usualy do is this routine:

  * manulay test connection to GRBL, jog a bit, zero run some g-code, do simple probing, test HW cycle start button
  * manualy test profile/pocket, tabs and cut. Sometimes also slice mesh and drag knife
  * upload new changes to git (and pypi) and wait few weeks for people to complain about problems

There are people using git version and they are super helpful with this. They helped me to figure out some major problems. Some bugs only happen on some platforms anyway. So i think it's enough to wait maybe month before releasing new version.

Harvie:
BTW regarding that "connection plugin"  infrastructure... I think i can do it. We only need to reimplement class with following methods:

```
__init__()
close()
flush()
flushInput()
inWaiting()
readline()
setDTR()
write()
```

AFAIK that's all that is used in sender. At least according to grep.

coddingtonbear:
Your approach toward testing & separating connection plugins into multiple classes each implementing a common interface -- I think that's definitely a great idea!  Separating out the serial "connection" into classes (x over tty; x over network; x faked for testing) is a great first step; on that same tangent: another possible step with wonderful benefits as far as maintainability is concerned would be abstracting each of the various gcode interpreters -- that is Grbl 1.1, Grbl 0.8 and others (cases like CarbideMotion) that each have their own special casing into their own classes.  That, though, is maybe not the right project to start on for now, mostly because it's not clear what that interface would look like.

About python 3 compatibility -- yeah; I do have quite a bit of experience working on moving projects from python 2 to python 3 and maintain a few libraries right now that support both simultaneously.  Unfortunately, the fact that bCNC does not have much in the way of tests makes that kind of a transition really a lot more difficult than it needs to be, and it'll be quite a slog to sort out all of the problems given that in most cases, you can only discover that a problem exists at runtime, and, as I'm sure you've noticed, one of the more troublesome things about bCNC as it that it's sort of a birds' nest of entangled rules, special cases, and multi-level menus.  I'm glad we're both talking about at least putting together these basic integration tests, though; because that'll at least make it possible!

As far as maintaining python 2 support alongside python 3 -- that's really mostly only important for libraries that are being imported by software running in a variety of environments, and although possible, it does definitely make the code a little more complicated.  Secondarily, the EOL of python 2 is really very close now (https://pythonclock.org/), so the benefits you'd get from maintaining python 2 support are even more diminished.  The way that most desktop applications (like bCNC) handle transitions like this is to declare that major version X and afterward will require python 3, and the vanishingly small (by this point in time) number of users who are trapped on python 2 for some reason must use only version X and below.  In case it's not clear, what I'd do if I were you is say that starting on v1.0 and onward, bCNC will require python 3.

I'm going to take another look at this PR sometime in the next week or so, but if you'd like to talk about strategies for moving for python 3 or need my help sorting out anything, you can reach me via e-mail, on freenode (I'm coddingtonbear there, too), or on google chat.  I'd love to help out where I can, but time is a little precious of course.

Harvie:
> another possible step with wonderful benefits as far as maintainability is concerned would be abstracting each of the various gcode interpreters -- that is Grbl 1.1, Grbl 0.8 and others (cases like CarbideMotion) that each have their own special casing into their own classes. That, though, is maybe not the right project to start on for now, mostly because it's not clear what that interface would look like.

Actually... As i said... i've already done this like week ago... #1020

>  one of the more troublesome things about bCNC as it that it's sort of a birds' nest of entangled rules, special cases, and multi-level menus.

Well... You only need to do transition from 2 to 3 once... Having tests might be nice, but you don't really NEED it. I believe that people will tests this manualy if needed.

> declare that major version X and afterward will require python 3

As far as i know, python3 is available almost everywhere with exception of Windows XP

> but if you'd like to talk about strategies for moving for python 3 or need my help sorting out anything, you can reach me via e-mail

Thanks! I always got stuck on something when i tried python3. I will surely send contact you if i ever try again.

coddingtonbear:
That's the challenge, actually!  How do you know when the 2->3 conversion is complete?  One of the major changes between Python 2 and Python 3 is that Python 3 requires much more careful handling of the distinction between bytes and strings; automated tools won't, in many cases, be able to help determine when you need one or the other or even if there's a problem, and those sorts of problems are very difficult to identify using strategies other than just exercising the code.  Were there tests, which are mostly useful in this scenario for ensuring that most lines of code are exercised, for a codebase of this size, it'd just be a matter of running `python-modernize` and spending a few hours fixing test failures when running the tests under Python 3 to end up with a codebase you could be reasonably sure was Python 3 compatible.

I'm of course willing to help out where I can, but I just want to make sure you understand what you're getting yourself into.

Harvie:
There are hidden problems in bCNC anyway :-) If the bug is not triggered by anyone after first month in git, i call it good enough to release :-)

Harvie:
I've just found that pyserial library already provides abstraction layer for serial connections.
That means we can write stuff like `socket://localhost:2217` to the serial port input box in bCNC, we just didn't realized it's possible. See #1072 for more info.

This kinda defeats the point of adding similar abstraction layer in bCNC. I think we should just implement braindead simple grbl simulator using SOCAT like grbl-sim does:

`socat -d-d PTY,raw,link=/tmp/ttyFAKE,echo=0 "EXEC:'./grbl_sim.exe 1 -n -s step.out -b block.out',pty,raw,echo=0"`

Harvie:
Also i've found that pyserial has method to attach your own serial handler, so you can use it to hook your fake grbl:
`protocol_handler_packages.append("my_handlers")`

Harvie:
I've got following travis setup working:

```yaml
language: python
python:
    - '2.7'
    - '3.7'
matrix:
    allow_failures:
        - python: '3.7'
install:
    - pip install -e .
script:
    - echo HOORAY
    - python -m compileall bCNC
    - python setup.py sdist
```

It does not test functionality, but it at least checks python syntax using compileall

coddingtonbear:
Awesome, thanks; I'm hoping to find a little time to spend on this next week sometime!

Harvie:
I've enabled strict check for inconsistent indentation.
Also i've added .travis config to v3 branch, which runs on python3 and is allowed to fail on python2 (opposite of what we have in master).
Travis now supports running jobs on windows. Currently they don't have python support ready. But once they figure this out (i guess it will happen relatively soon), we will be able to use travis to automaticaly build and upload .exe files for windows users.

Harvie:
Rudimentary Travis tests are in place. I will close this PR in favor of #1117 issue, where i want to discuss the possibilities further...

Issue #595, opened 2017-05-20
An issue with probing


Avalonnw:
I noticed that probing parameters (I tried speed, tool change coords and probing coords) won't apply until software restart. Is that supposed to work like that?

vlachoudis:
Nope. They should have an immediate effect. Watch the output in the terminal for any traceback message.

Avalonnw:
No output there. I will try to monitor this behavior tonight.

eyal0:
This can happen if you change the parameters without pushing the tool change or calibrate button.  For example:

1. Set feed to 1000.
2. Push calibrate. It will work at 1000.
3. Set speed to 10.
4. Push play on gcode that has M6 in it.
5. When M6 is encountered, feed rate will still be 1000.

That has been my experience. It is because the probeUpdate only happens when pushing calibrate or tool change and not from M6 tool change command. Bummer.

eyal0:
Try https://github.com/eyal0/bCNC/tree/probe_update_on_change if you like.  It has a fix.

The master branch also has a feature to allow you to have probe with decreasing speed (first probe at speed 1000, then 100, then 10).  Same accuracy but faster tool change.

vlachoudis:
@eyal0 thanks for spotting. I will check both patches on my machine

vlachoudis:
Thanks

Avalonnw:
I tried the new version and had to go back to old one.
The new version does probing, then moves off a bit and then goes to alarm state saying the travel requested exceeds machine limits.
It might be related to my other issue with tool change mid-program as after probing I get the same message. I take it's something to do with GRBL itself (rounding problem?) but have no idea how to rid of this problem. If anyone could help, I would be grateful!

eyal0:
@Avalonnw Can you check the terminal output and see what commands are being run?  Maybe you set the probe start or probe distance poorly and the tool is going out of bounds during the probing?

eyal0:
@vlachoudis I see that https://github.com/eyal0/bCNC/tree/probe_update_on_change was never merged.  Maybe it's not needed anymore?  I will check using the [steps listed above](https://github.com/vlachoudis/bCNC/issues/595#issuecomment-303936858) and let you know.

eyal0:
I just tested on my CNC router and the bug is still there.  I'll resend that PR.

Avalonnw:
[...]

PR #706, opened 2017-12-10
Probe update on change

eyal0:
The probing speed should update as soon as the user changes, not only when pressing calibrate.  This is important because M6 (tool change) commands in the gcode file should use whatever the user most recently typed, not whatever was in the text box back when calibrate was pressed.

This fixes the problem by updating the probe feed speed as soon as the textbox is changed.


eyal0:
Fixes https://github.com/vlachoudis/bCNC/issues/595

vlachoudis:
Thanks

PR #725, opened 2018-01-19
Update bCNC.pot [root]

SteveMoto:
state: "POT-Creation-Date: 2018-01-19 18:53+0100\n"


vlachoudis:
Thanks I've updated the pot externally

PR #730, opened 2018-01-26
 update fast probe feed when the field is edited

goddisignz:
Fast probe feed was only updated when the "regular" probe feed was modified. Now changing the fast probe field also calls the updateProbe()-function.


vlachoudis:
Fixing to z-1,z1 is not correct. You are making an assumption that 1 "unit" is ok. I  can see immediately a problem with the imperial units. It had to be dynamically calculated.

goddisignz:
Interesting, the last commit should not be there. Only commit e531bc5 is meant for pulling. Should I create a new request?

vlachoudis:
Sorry, I forgot to push

PR #731, opened 2018-01-27
changed probing behaviour to avoid exceeding the soft limits

goddisignz:
Changed probing behaviour to avoid exceeding the soft limits during tool change and calibration.
Problem:
During M6 (manual TLO/WCS) commands or the probe is lowered (g38.2) by the tooldistance. If fastprobefeed is set it is raised with the g38.4 and lowered again (g38.2) by the tooldistance. As the tool does not move far from the touch plate the second lowering can exceed the soft limits althoug they are never reached.
Solution:
Probe only 1mm after the probe lost contact.
Optional other solution:
Do not reverse probe but only retract for example 1mm and then probe 1.5mm or so.

### DIFF
--- a/CNC.py
+++ b/CNC.py
@@ -1681,18 +1681,26 @@ class CNC:
 			lines.append("g53 g0 z[toolprobez]")
 
 			# fixed WCS
-			if CNC.vars["fastprbfeed"]:
+			# probe multiple times with decreasing probe feed
+			if CNC.vars["fastprbfeed"] and CNC.vars["fastprbfeed"] > CNC.vars["prbfeed"]:
 				prb_reverse = {"2": "4", "3": "5", "4": "2", "5": "3"}
-				CNC.vars["prbcmdreverse"] = (CNC.vars["prbcmd"][:-1] +
-							     prb_reverse[CNC.vars["prbcmd"][-1]])
+				CNC.vars["prbcmdreverse"] = (CNC.vars["prbcmd"][:-1] + prb_reverse[CNC.vars["prbcmd"][-1]])
 				currentFeedrate = CNC.vars["fastprbfeed"]
-				while currentFeedrate > CNC.vars["prbfeed"]:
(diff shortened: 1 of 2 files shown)


vlachoudis:
Retracting maybe is "safer" than reverse probing. But still I am not confident on a fixed distance.
Here the "retract" is the tooldistance which can easily go out of the limits

goddisignz:
Ok, i see the problem. I will give it another try and calculate the probe distance by the current machine coordinates, the current working coordinates, the probing location and the tooldistance.

vlachoudis:
I think I see the problem. I was naively converting the tooldistance position to absolute by adding the wz-mz. Instead I need to reverse/probe either to the absolute of toolprobez distance but converted to work coords, or relative the difference from the toolprobez-mz.
I've uploaded the changes on the new master. Can you test it please?

goddisignz:
I see the problem at https://github.com/vlachoudis/bCNC/blob/493dab44518d8678b80540b246a2331445348217/ProbePage.py#L1698 and https://github.com/vlachoudis/bCNC/blob/493dab44518d8678b80540b246a2331445348217/ProbePage.py#L1703
The retract/revprobe is not the problem because the given distance is never travelled. You stop right above the touch probe. If you now try to move down by tooldistance the resulting position is/may be out of the machine travel limits although this position is also never reached.
As I said, I will come up with a clean solution that does not require fixed numbers.
Btw. I can not find any changes with respect to this problem on the current master.

vlachoudis:
Check the latest commit.

goddisignz:
Looks better, but the problem with the z[-tooldistance] remains.
Have a look at https://github.com/goddisignz/bCNC/commit/6a64977f4a45183225f0241d349c405d7302936e
This should do the trick as the probe target position is always toolprobez - tooldistance.

vlachoudis:
I've implemented your suggestion. I didn't know how to merge only this specific commit, so I copied the liens. Could you please check

PR #733, opened 2018-01-29
added "%wait" commands to obtain the correct mz during probing

goddisignz:
I made wrong assumption about the "mz". I thought it is evaluated after the previous command but in fact it is evaluated as soon as it is fed into the command queue and thus, leading to wrong calculations.
The "%wait" now ensures that the movement has finished before evaluating mz.
Now all my probing scenarios work correctly.


vlachoudis:
Indeed the m[xyz], w[xyz] are asynchronous.

goddisignz:
BTW, the implementation without the "%wait" does not work. It will still generate errors.

PR #742, opened 2018-02-08
Update Chinese translation

Georgwei:


Georgwei:
Because of the large number of Chinese users, I tried to translate it into
Chinese. In the future，There will be some need to modify the content.
I am not very familiar with Python and Github, the operation may be some
mistakes. There will be something wrong with running bCNC.
I don't know how to solve:
1. There is no "Simplified Chinese" option in the language section;


2. Chinese can not all be translated;

​
3. Exit, an error occurred.

​
4.Python command bar will be displayed at run time

2018-02-08 17:08 GMT+08:00 Vasilis Vlachoudis <notifications@github.com>:

vlachoudis:
1. There is "traditional chinese". Should be renamed to "Simplified Chinese?" or it is different?
2. What cannot be translated can remain in English
3. I would need more information, what error etc..
4. In windows, the python terminal appears all the time. You can change in the .bat the interpreter to pythonw to avoid the terminal. However the terminal can potentially provide useful information in case of errors

Georgwei:
1.Simplified Chinese is a modern and improved text which commonly used in
mainland China. Traditional Chinese is commonly used in Chinese in Taiwan,
Hong Kong, Singapore or other Chinese outside of China. The two texts are
different, so an option needs to be added.
2.As shown, Traditional Chinese can be all translated, while simplified
Chinese can not? Is bCNC.ini problem?


3.​When exiting bCNC (click to exit), you can not exit with an error
message（as shown）

​

2018-02-08 18:49 GMT+08:00 Vasilis Vlachoudis <notifications@github.com>:

Georgwei:
Thanks
After re-translation, 1-2 problems are solved, but 3. problems still exist

​

2018-02-09 15:46 GMT+08:00 Vasilis Vlachoudis <notifications@github.com>:

PR #746, opened 2018-02-14
synchronize zh_CN .po and .mo files with bCNC.pot

SteveMoto:
bCNC.pot [POT-Creation-Date: 2018-01-29 09:17+0100]


vlachoudis:
Thanks

vlachoudis:
I've added an entry as "Simplified Chinese". Is it correct?

PR #785, opened 2018-03-24
Support single stroke fonts

Quiigi:
This change allows suppressing spurious closing lines when drawing text in a single-line font.

In [Summary of single-line fonts](http://imajeenyus.com/computer/20150110_single_line_fonts/index.shtml), imajeenyus writes: "Single-line fonts (also called engraving fonts, technical lettering fonts, pen plotter fonts) are hugely useful when performing machining or drawing operations."  Under "Notes/warnings", that page mentions the issue of "closing lines".

True-type fonts are often outline fonts, and some software always adds closing lines, which the page above highlights in red.  This proposed change adds a boolean parameter "Closed" to the "text" plugin.  It's default is _True_, maintaining the current behavior.  But with a single stroke font like MechSoft from [McNeel](https://wiki.mcneel.com/rhino/engravingfonts), you want to turn this off.  It saves the extra step of deleting the spurious lines manually.


Quiigi:
Example:
single-stroke

vlachoudis:
Thanks

PR #798, opened 2018-04-10
Driller use anchor and name in Editor

buschhardt:
Hallo,
I insert a option to use only the anchor points for drillholes.
Second I use the name for the name in the editor.
bildschirmfoto von 2018-04-10 12-28-37
bildschirmfoto von 2018-04-10 12-29-07


vlachoudis:
Thanks

Issue #809, opened 2018-04-19
Drill with excellon


buschhardt:
Hallo, I have modified the driller plugin to import excellon drill files. Its only a try to see if its work an what you think about.
I get the drill file from [Fritzing](http://fritzing.org) when you export extended Gerber, there is a drill file (filename_drill.txt), that file I import in the driller plugin. The values in the () is the diameter of the hole in the system unit.
I attach the driller-plugin an a testfile.
So what you think - is it needed?

![bildschirmfoto von 2018-04-19 22-27-19](https://user-images.githubusercontent.com/1921078/39016515-33380448-4410-11e8-8a51-b8685bd2efaa.png)
![bildschirmfoto von 2018-04-19 22-29-34](https://user-images.githubusercontent.com/1921078/39016584-6668507a-4410-11e8-87fc-3bdf0b5a7f4c.png)
![bildschirmfoto von 2018-04-19 22-30-19](https://user-images.githubusercontent.com/1921078/39016614-7eeae61c-4410-11e8-9b70-966df2eaae83.png)

[drillexcellon.zip](https://github.com/vlachoudis/bCNC/files/1929681/drillexcellon.zip)

buschhardt:
OK, thank you - I will do a pull request. Maybe a wiki - topic helps too.

PR #827, opened 2018-05-09
Excellon drill file import

buschhardt:
See https://github.com/vlachoudis/bCNC/issues/809


buschhardt:
Here is a sample from fritzing export: 
[drill.txt](https://github.com/vlachoudis/bCNC/files/1989124/drill.txt)

PR #831, opened 2018-05-11
Helical cutting

MARIOBASZ:
…lock

I have made some improvements in my Helical plugin.
Performance: Milling / threading right / left, internal / external.
Choose diameter and pitch of the helicoid.
Choose the end mill (or use the last one selected if not chosen)
Choose the starting location of the helix from the panel or from where the selected block starts (it would be better than in this option, the helicoid is added at the beginning in the same one instead of in a new one, my knowledge of Python still they are insufficient, I have based on Filippo Rivato's Driller plugin and Buschhardt's improvement, if someone wants to implement it, thanks, or as a new option in the control panel, include in the block / generate new block
Choose the starting and ending heights,which allows you to mill / tap anywhere on the material, not only at the end.
Set the safety margin from the lateral edge at the input and output of the tool.
Milling / tapping at a height higher than the initial one (I do not know if it has any use).
Eliminated passed in initial and final levels for threading, only remains in type of helical cut.

Interesting improvement if someone wants to add it: To make a thread of fillets wider than the height of the T type end mill, it would be necessary to generate copies displaced in height for each successive pass.
Happy with the result, tell me if there are errors and I hope it is useful.
Thanks again Viachoudis for your software, to those who collaborate, especially to Rivatto and Bruschhardt for your work and Harvie for your observations (they really helped me)
Greetings, Mario


vlachoudis:
Thanks. Just one comment please do not add the .pyc files in the repo

PR #833, opened 2018-05-12
Driller closed path

buschhardt:
If the path closed (last point == first point) or a point is multiple times in the process list, it will be remove from list.


vlachoudis:
thanks

Issue #590, opened 2017-05-14
Add support for helical cutting / ramp plunging


Harvie:
In lots of cases it's much better practice to plunge down along with cutting in X/Y axes, so you prevent mill from going straight down. For some tools it's even the only option... Can you please add this as option for CUT operation?

Eg. when you cutting circle. You don't go down in Z and then do XY circle and then go down in Z again. Instead you can do helix cut down with last layer being the only flat one.

![image](https://cloud.githubusercontent.com/assets/276504/26034759/d6640f00-38c1-11e7-8d57-7035375c3fcf.png)

Fosforitofercho:
It is a feature of your g-code generator, Use Fusion360, has a LOT!!!! of features like this

Harvie:
@Fosforitoferch I hope i will see this feature in bCNC. Fusion360 does not work on Linux. Also i guess this is quite OT.

Harvie:
It seems to me, that related code is located in `./CNC.py` file in following methods:
```
def fromPath(
def cutPath(
def cut(
```

Harvie:
I am not really python guy and i don't really understand inner workings of bCNC, but i've tried to modify these two functions and get some results. This is image of helicaly cut rectangle. I just selected rectangle from DXF and clicked on CUT button in bCNC:

![image](https://user-images.githubusercontent.com/276504/41948968-9a1085ac-79bf-11e8-8747-4e02a85177c9.png)

Here is the code:

fromPath()

```python
	#----------------------------------------------------------------------
	# create a block from Path
	#----------------------------------------------------------------------
	def fromPath(self, path, block=None, z=None, entry=True, exit=True, zinit=0):
		if block is None:
			if isinstance(path, Path):
				block = Block(path.name)
			else:
				block = Block(path[0].name)

		def addSegment(segment, z=None):
			x,y = segment.B
			if segment.type == Segment.LINE:
				x,y = segment.B
				#block.append("g1 %s %s"%(self.fmt("x",x,7),self.fmt("y",y,7)))
				if z is None: block.append("g1 %s %s"%(self.fmt("x",x,7),self.fmt("y",y,7)))
				else: block.append("g1 %s %s %s"%(self.fmt("x",x,7),self.fmt("y",y,7),self.fmt("z",z,7)))

			elif segment.type in (Segment.CW, Segment.CCW):
				ij = segment.C - segment.A
				if abs(ij[0])<1e-5: ij[0] = 0.
[...]

PR #887, opened 2018-06-29
Added real ramp/helical cutting functionality, see #590, and more...

Harvie:
Added real ramp/helical cutting functionality, see #590

image

image


Harvie:
I've also added little checkbox to create pocket and profile in single step:

image

PR #890, opened 2018-07-01
new screenshots [wiki] pt. 2

SteveMoto:
sources: https://github.com/SteveMoto/bCNC-dev


SteveMoto:
https://github.com/vlachoudis/bCNC/wiki/Customize-DRO

Image 3 of 3 still needs to be corrected, marker must be 5 lines down. Will be redelivered ...

Issue #897, opened 2018-07-02
Can't properly detect path direction


Harvie:
I made separate issue from this, as this affects multiple other issues: Namely #881 and #840 

Here you can clearly see that in some cases it gives oposite direction (see segment direction vs. overal path direction):

![image](https://user-images.githubusercontent.com/276504/42172240-bc190498-7e1b-11e8-8251-63c095eef917.png)

When it fails to properly detect path direction i cannot properly implement conventional/climb milling. Also when i try to do pocket operation on such wrongly identified rectangle with overcuts it crashes on ifinite recursion OR shows this:

![image](https://user-images.githubusercontent.com/276504/42172073-3ea9c024-7e1b-11e8-9ada-be675adcfa0e.png)

Harvie:
I've reimplemented the algorithm using stackowerflow answer with best rating:

https://stackoverflow.com/questions/1165647/how-to-determine-if-a-list-of-polygon-points-are-in-clockwise-order

So far it seems to work:

![image](https://user-images.githubusercontent.com/276504/42176873-c4a82c38-7e2a-11e8-8d69-d356fe4d2008.png)

I've done some hacks to make it work on paths that are mostly arcs, but it can be relatively simply improved if needed.

Harvie:
Seems to be fixed fo 99.9% of cases

Issue #881, opened 2018-06-27
Climb vs Conventional milling is other way round in bCNC! Defaults probably not sane.


Harvie:
Right now it's:

Conventional = Always CCW
Climb = Always CW

bCNC has got this wrong. It should be like this:

Conventional = CW for inside profiles and pockets, CCW for outside profiles
Climb = CCW for inside profiles and pockets, CW for outside profiles

If you don't want to fix this, please at least rename it to CW and CCW, since climb and conventional is different idea.

Also i think conventional should be default! (as most people using bCNC have hobby machines which suffer from deflection). That means that inside profiles and pockets should be CW by default and outside profile should be CCW by default.

I use bCNC on shapeoko and cutting 5mm slot pocket with default settings resulted in 4.2mm slot (because of 0.4mm deflection), so my bolt does not fit to it. When i manualy changed to use real "conventional" milling (by clicking climb!!!), the slot is exactly 5mm wide. This can be confusing for beginners with cheap machines. Sometimes it makes sense to use climb milling to improve surface finish, but i believe the dimensional accuracy is more important in most cases. And people who will need to improve surface finish need to tweak more settings anyway...

Harvie:
Just saw the code:
`_direction()` returns 1 or -1 depending on if the path is CW or CCW
`cutDirection()` then reverses the path if direction is different than what user specified

This means there's no way for the code to really know if it's climb or conventional milling.

I think we should:
0.) **Stop calling this feature "conventional" and "climb"**
1.) Turn all paths imported from DXF to CCW
2.) Turn all inside profiles and pockets to CW
3.) Turn all outside profiles to CCW
4.) Rethink how to add real conventional and climb functionality on top of this

Harvie:
I just made this little more straightforward by adding CW and CCW to DIRECTION command (see the pull request). I used 1 and -1 values for this. Climb and conventional then got values 2 and -2, which are temporarily redirected back to 1 and -1 for now. But it's now clear what is the user's intent. So we can handle climb/conventional differently than cw/ccw to get appropriate behaviour.

I need following to implement real climb/conventional:

inside cutDirection() i need way to read these highlighted tags:
![image](https://user-images.githubusercontent.com/276504/42139818-7bbf72b4-7d95-11e8-8fce-098d2803fd76.png)

also i need way to show error popup to user if such tag is not present.

these are two things that i didn't figured out yet. any ideas?

Harvie:
[...]

Issue #815, opened 2018-04-27
Adaptive clearing for pocket operation


Harvie:
Please add feature to mill pockets using adaptive toolpaths. It helps to reduce deflection and allows to make deeper cuts (=tools will last longer, because you can use full length of blade, not only end of it) by keeping constant engagement in XY direction (could say "constant stepover", but it's more complex than plain stepover approach).

![image](https://user-images.githubusercontent.com/276504/39362163-ae212ee4-4a25-11e8-967f-85484e4e8956.png)

Avalonnw:
You realize there are CAM processors out there that do that, right?

Harvie:
@Avalonnw Exactly. That's why i think bcnc can do it too. Every time i post issue about CAM functionality some fusion 360 fan pops out of nowhere. Why even bother watching bcnc? Pretty much everything can be done by different software packages. With this kind of attitude we can all just go home and use something else. But this is bcnc repository, so i don't care if solidworks already does this.

Avalonnw:
Fair enough. I personally don't use bCNC for it's CAM as it's generally very confusing than any other option available. To me it's a nice grbl controller that can do many things others can not.

Harvie:
bcnc is extremely handy 2D CAM once you unconfuse yourself. not really for complex 3D parts, but i always use it for quick 2D jobs, because the workflow is much faster in such cases.

vlachoudis:
@Harvie I tend to agree with you (not because I wrote it :) but for simple 2D operations or almost real-time control make the CNC machine behaving like a sophisticated hand-tool, and it is faster than all other solutions. For complex things I use fusion 360, but there like all other packages you have to prepare everything in the program simulate it and then transfer to the machine.
In bCNC the CAM algorithms especially the pocketing algorithm, as well the workflow needs to be improved. My only problem is the lack of time. If someone is willing to sketch the algorithm it would be faster to implement.

Harvie:
This guy looks like he can help. But i guess, that we have to first implement way to work with differences of shapes. Since it does not make much sense to have 2D adaptive, but not have such basic feature as islands in pockets.

http://www.freesteel.co.uk/wpblog/faq/

Also gave me interesting idea about "trochoidal machining", that's similar concept, not as cool, but WAAAAY easier to implement.

Harvie:
I've written some elementary and highly experimental trochodial plugin:

[...]

Issue #220, opened 2016-01-31
How to pocket the difference of two shapes


cclein:
I'm wracking my brain trying to figure out how to do the following:

Say I have two circles or squares, one inside the other. Is there a way for me to say pocket the inside shape and then do the difference between the outside shape and the inside shape to a different height? Specifically lower.

It was quite easy to do in JScut and I'm wondering if I'm missing something in bCNC.

vlachoudis:
No you are not missing anything. For the moment is not possible to do. Is in my TODO list :)

cclein:
OK Good. I'm really loving your software and the rapid development!

I'm going to try and create the gcode on jscut.org. Do you have any
tips on importing that into bCNC? I tried a long time ago and had
issues.

Cheers,

Chris

vlachoudis:
It importing doesn't work (strange) please send us the gcode so I could debug the problem.

cclein:
Actually importing worked much better than the last time I did it.
Unfortunately none of the blocks etc for each operation came over but I
could live with that. Bad news the code was too big for my Raspberry Pi :(

vlachoudis:
@cclein could you send me the input to check where it failed?

cclein:
It didn't fail. Just too large for the RaspberryPi. Became unresponsive and
slowed cutting down. The file is 10mb. I can't send it via github. Anywhere
else I could send it?

vlachoudis:
Can you zip it and send it via email. My address in the about of the program. There are a few things I want to optimize

vlachoudis:
@motanescu interesting idea.

Harvie:
This would truly be usefull feature to have.

@motanescu I tried to make circular pocket with nested circular island like this:

![image](https://user-images.githubusercontent.com/276504/39668962-1f212e82-50de-11e8-80f2-afcfebba3d14.png)

But the pocket failed to generate

```
Traceback (most recent call last):
  File "/opt/bcnc-git/Utils.py", line 413, in __call__
    return self.func(*args)
  File "/opt/bcnc-git/ToolsPage.py", line 1490, in execute
    self.tools.getActive().execute(self.app)
  File "/opt/bcnc-git/ToolsPage.py", line 769, in execute
    app.pocket(name)
  File "/opt/bcnc-git/bCNC.py", line 1758, in pocket
    msg = self.gcode.pocket(blocks, diameter, stepover, name)
  File "/opt/bcnc-git/CNC.py", line 3637, in pocket
[...]

PR #898, opened 2018-07-02
Real conventional/climb milling. Probably better direction() algorithm. Trochoidal plugin! Difference plugin.

Harvie:
See #897 , #881, #815, #220

I also added alpha version of trochoidal plugin, so we can start experimenting:

image

And alpha version of shape intersection plugin (technology preview for pockets with islands):

image


vlachoudis:
Fantastic. Many thanks!

Issue #903, opened 2018-07-03
Rounding problems


Harvie:
When i generate inside profile of rectangle with overcuts and then try to do outside profile of this profile it leaves out segments around overcuts due to fact these are not rounded properly. When i manualy round the line it gets fixed. Also when i use Editor -> Edit -> Round tool.

![image](https://user-images.githubusercontent.com/276504/42215968-7f7c5152-7ec0-11e8-8981-7f1e4abec73e.png)

I would fix this, but i am not sure what is the bCNC's policy on decimal precision and where is good place to round it (when g-code is generated or when it's read?).

Harvie:
This fix caused different problem:

![image](https://user-images.githubusercontent.com/276504/42419038-49046f24-82ad-11e8-8292-fc2c84779a7a.png)

Harvie:
To sum the problem:
  - when i don't round the blocks, some segments are missing from profile
  - when i round the blocks some arcs are repeated twice in profile

@vlachoudis any ideas pls?

Harvie:
it was working at 6b468a573e18f47d473fb8645a6366ef5456defc
and broken at 868b04dad169cabf76740cf5a0c48ffe170851dd
so the bug must have been introduced somewhere between...

Harvie:
Can you please reopen this? It was not really fixed by my commit. i reverted that later. We have to sort this out...

vlachoudis:
Sorry it was by closed by mistake. I will try to reproduce and see what is going on.

Harvie:
Thank you. Here is the test file for the second case:

```
(Block-name: Header)
(Block-expand: 1)
(Block-enable: 1)
M3 S12000
G4 P3
(Block-name: 0)
(Block-expand: 1)
(Block-enable: 1)
G0 X5.257 Y8.5573 Z0
(entered)
G1 Z0 f200
G1 X7.2021 Y-13.2064
G1 X-8.4555 Y0.1244 Z0
G2 X-4.7178 Y9.2661 I3.3693 J3.9574 Z0
G1 X-2.7228 Y9.1244 Z0
G3 X1.2671 Y8.8409 I1.995 J-0.1418 Z0
G1 X5.257 Y8.5573 Z0
(exiting)
G0 Z10
(Block-name: Footer)
(Block-expand: 0)
(Block-enable: 1)
M5
G0 Z20
```

[...]

PR #906, opened 2018-07-04
Pocket islands! Custom shaped tabs! Improved trochoid! Rounding problems, Island tags!

### LINKED ISSUE #767: Disabled blocks in saved g-code still interpreted by other programs
When i save gcode in bcnc, even the disabled blocks are saved. That makes sense.
But when i want to use this gcode in 3rd party software it can't tell disabled blocks by comment header and just runs it as if it was enabled.

Can you please put all G code commands in disabled block into braces (or use whatever g-code syntax is suitable to comment out inactive parts of g-code file)? So it does not get interpreted by 3rd party software like universal g-code sender. This causes strange behavior. (my mill following original drawings before doing actual cutting)

Harvie:
Fixed #767
Probably solved rounding problem from #903
Just added way to tag path as an island (also sets color to red) for future purposes of #220:

image

Aaaand custom tabs are here!

image

You can use them to make islands! :-)

image

image


MARIOBASZ:
I congratulate you for your work! I'll look at it more closely. I am also working on my version for trochoidal mode. I am interested in your idea of ​​islands, I hope that you go up to the main frog to download it, or is there a way to prove it? Regards

Harvie:
@MARIOBASZ of course, it's in my repo. that's how i made this pull request:
https://github.com/Harvie/bCNC

Harvie:
@MARIOBASZ are you going to implement trochoidal pocketing? it's surely easy to implement on simple grooving/cuting operation. But doing real trochoidal pocket is completely different chalenge. I will be modiffing the pocket code to make it obey islands and produce better toolpaths. See #815 and #220 for details. If you are going to modify the pocket code. So it does not have to be rewriten to support the islands...

Harvie:
commit f23d0d002bb7794df5c0617592ab633affc27144 (Probably solved rounding problem from #903) seems to break the profile operation in some cases... i'll try to fix it

Harvie:
I originaly thought that i caused #903 bug by some of the commits in this PR, but it seems it was present long before my first contribution to bCNC. So i would call this pull request ready to merge.

Anyway. I'll try to later bisect where that happend, since #903 does not exist in stable release.

vlachoudis:
Many thanks.

MARIOBASZ:
@Harvie , my version of the trocoidal plugin I developed it from the Driller plugin. My knowledge of python is very rudimentary, I use few usable libraries (I step reinventing the wheel ..), I will definitely use your plugin, it is much cleaner.
The use of the trochoidal cut is surprising: I have machined aluminum in 4mm passes when before I could only make passes of 0.1mm. Ingeniously simulate adaptive cuts by doing pocket operations in trochoidal mode, changing the length and modifying the diameters of the trocoid in the overcutt to reach the angles of the silhouette with the diameter of the bit instead of that of the trochoid. Regards

Harvie:
> My knowledge of python is very rudimentary

I only learned python like week or two ago from bCNC code :-D i didn't even understood the g-code back then. Your helical plugin inspired me to learn how arcs are made in g-code and then commits started to appear :-)

> I have machined aluminum in 4mm passes when before I could only make passes of 0.1mm.

You did that with bCNC? Do you have any videos of machine using trochoidal code generated by bCNC? Do you use clockwise or counterclockwise trochoid loops?

> Ingeniously simulate adaptive cuts by doing pocket operations in trochoidal mode

Yes, that's what you can do. Fool profile/pocket operation to think you have bigger endmill and then apply trochoidal postprocessor on resulting toolpath.

> modifying the diameters of the trocoid in the overcut to reach the angles of the silhouette with the diameter of the bit instead of that of the trochoid

Cool idea! Do you have code for that? BTW this is very similar idea to v-carving (which is also machining with variable diameter). I guess we can create code for v-carving and then reuse it for trochoids :-)

I would also like to make trochoid loops shorter to speed up the machining. right now it's not very efficient.

MARIOBASZ:
1-I'm glad I helped you with your help, your comments on my helicodal code helped me improve it. In fact, I have to change changes or rewrite and base on your trochoidal: the most important is to select different silhouettes to perform helicoids.

2- Yes, with BCNC. My actal machine is very little solid (structure of MDF, GT2 belts, without rectified guides .. I am putting together a more serious one, spindle 500W) .. it should not be able to mechanise aluminum,. I have not experienced yet unlike G2 and G3 in trochoid

3- My solution was to define a trochovercut, First I do a trochoidal route profile operation with an overcut and adaptive profile option, there is a policeman to indicate that the overcutt starts so that it can modify the diameters. If you find it useful I can move on or implemented / modified
Carving in V you mean to make a standard or inverted cone?
It is one of the variations that I plan to make in Helical, separating it into cut, threaded and cone.

4-I have optimized like you joining each trochoid with the next in a straight, circular or arc way. The solution is to increase the acceleration, for this I think it will serve to use Tynig2 to avoid the lashes

I reasoned that to modify the diameters of trochoids the ideal is to work with vectors of 4 axes (x, y, z, dTroch). It was not my method.
To add to your code, which I have implemented in mine, is to return in a straight line when returning back through the same route (for example in the overcut) .. or return with trochoids of zero diameter,
It also allowed trochoidal mode in 3D.
I will send code example as Issue.
Vlachoudis, Harvie, a pleasure




Abrir en Google Traductor | Comentarios
-- | --

Harvie:
> To add to your code, which I have implemented in mine, is to return in a straight line when returning back 

I used to do this, but it's not good idea. You should ease to the corners to prevent machine from suddenly changing movements. The path should be as smooth as possible. No sharp corners! I am planning to use spiral path rather than circular for return. Also using eliptical cuts rather than circulars can shorten the distance that's needed to be travelled in single pass.

MARIOBASZ:
with straight line return or zero radius trochoid I was referring to the return path after overcut,

Harvie:
>  I think it's good to let the speed management and acceleration (and jerk) leave it to the (_controler_).

i don't think so. if you leave it to the controler it will slow you down because of acceleration an jerk control. that's why it's better to do round cuts. that's the point of trochoidal. if you were cuting in flat layers (squares) instead of trochoids it would cause machine to slow down (in adition to being harsh on cutter). It's basic idea of trochoidal machining and similar strategies to never change direction suddenly, so controler is already accelerated when going out of corner.

MARIOBASZ:
Maybe it will inspire you to determine the trajectory of the spirograph plugin.
I will continue the conversation from your open request pul # 911, more people can see and contribute concerns.
A hug, I reiterate, great work.

justinclift:
> i don't think so. if you leave it to the controler it will slow you down because of acceleration an jerk control.

Doesn't that depend on the controller look ahead + algorithms used for motion planning?

For example, g2core has a [much better motion planner](https://github.com/synthetos/g2/wiki/Jerk-Controlled-Motion-Explained) than the standard one currently in GRBL.

That being said, the GRBL one will probably get improved some day too. :smile:

Harvie:
> Doesn't that depend on the controller look ahead + algorithms used for motion planning

It does not. If your motor is turning one way and suddenly it has to turn the other way. obviously you have to stop/decelerate and accelerate again in other direction. unless you change your path, which is something that cannot be done by motion planner.

Issue #901, opened 2018-07-02
Slicing STL


Harvie:
There's STL slicer writter in python that would probably be good enough for integration into bCNC:

https://github.com/matthewelse/pySlice

All we need is to extract 2D outlines (layer cross cuts) from STL, so we can then use bCNC to create operations on such outlines.

Harvie:
There's very impressive python library called PyMesh:

https://pymesh.readthedocs.io/en/latest/index.html
https://github.com/qnzhou/PyMesh

It can Read/write 2D and 3D mesh in .obj, .off, .ply, .stl, .mesh (MEDIT), .msh (Gmsh) and .node/.face/.ele (Tetgen) formats.

https://pymesh.readthedocs.io/en/latest/basic.html#loading-mesh

It has simple API, that returns object containing paths for individual slices:

https://pymesh.readthedocs.io/en/latest/api_geometry_processing.html#slicing-mesh

I guess we can have basic 2D slicing in like 10 lines of code...

Harvie:
Another possibility: https://github.com/julienr/meshcut

Harvie:
I succesfully sliced mesh to g-code using meshcut library!

![image](https://user-images.githubusercontent.com/276504/42540876-638108a4-84a0-11e8-8c25-df47c0ecace0.png)

PoC code:

```python
"""
Basic usage example on a ply mesh. Note that this require a closed, manifold
input mesh.
"""
##
import os
import numpy as np
import itertools
import utils
import ply

import meshcut

if __name__ == '__main__':
    example_fname = "/home/harvie/Temp/meshcut/examples/data/mesh.ply"
    with open(example_fname) as f:
        verts, faces, _ = ply.load_ply(f)

    plane_orig = (0.02, 0, 0) #vyska
    plane_norm = (1, 0, 0)

    contours = meshcut.cross_section(verts, faces, plane_orig, plane_norm)

    for contour in contours:
        print("(---------------)")
        #print(contour)
	first = contour[0]
	print("g0 x%f y%f"%(first[1],first[2]))
	for segment in contour:
		print("g1 x%f y%f"%(segment[1],segment[2]))
	print("g1 x%f y%f"%(first[1],first[2]))
```

Harvie:
I've got first version of plugin!
![image](https://user-images.githubusercontent.com/276504/42542702-65f4d748-84a8-11e8-9f61-7de6daae3f74.png)

Harvie:
[...]

PR #911, opened 2018-07-08
3D mesh slicing! Improvements on islands, Tabs are now editable islands too, CAMotics compatibility, Major trochoid improvements, Join/Split blocks!, Fixed move of G91

### LINKED ISSUE #915: G91 problems when moving or otherwise modifying
When having block like this utilizing G91 it will fall apart when moved:

```gcode
(Block-name: block)
(Block-expand: 1)
(Block-enable: 1)
g0 x0 y0
g1 z0
g1 x10 y0
g91
g1 x0 y10
g1 x-10 y0
g90
g1 x0 y0
g0 z10
```

This is kinda fun, because you can make blocks that will change shapes when moved. However it's not really practical for CNC use:

image

This is due to the fact that coordinates between G91 and G90 are translated too. bCNC should stop modifying coordinates after G91 and resume after G90 when moving block.

I am not really sure if this bug also affects other operations.

BTW Fixing this might make hand editing of tabs even more user friendly, since it will be needed to edit only 1 segment instead of 4 if we'll use G91 to describe tabs.

Harvie:
Improvements from https://github.com/harvie/bCNC

image

I simplified the use of islands. now you don't need to create tab object from islands, you can directly cut... #220 

image

image

Smoother trochoidal mode #815  Also supports arcs now :-)

image

Tabs are now implemented using islands #912 

image

Trochoidal in combination with tabs and camotics :-)

image

Join blocks together and split them again :-) Useful for modifiing individual tabs! #912 
Fixed #915 

image

Optional circular tabs to get consistent tab size over lines with various angles

image

image

New plugin can slice 3D meshes in STL and ASCII PLY format! #901 
Now we have two modes.

image

Non flat:

image

Flat:

image


MARIOBASZ:
Hi, I have detected this problem when trying to use islands in pocket operation: sometimes it cuts inside the island instead of keeping Z safe

image
Here it happens
image
Not here
image

I think it would be good if Z is safe to move with G0 (I only made a partial modification when entering the island)

MARIOBASZ:
If it is viable, it would be that you ignore the movements (ectors or points) when you are traveling within the island
Another idea: you can make a union of islands, and that results only the perimeter of the unions, but eliminating the intersection of them (what resuta of your operation difference)
We would be one step closer to make a more efficient PCB routing

MARIOBASZ:
I have seen that in the implementation already uploaded to the master branch, in the cut operation the heights in and out of tabs do not work well.
I leave the question here or do I upload it as a neva issue?

Harvie:
@MARIOBASZ it seems to me that you do not use latest version from my fork, i've already fixed these issues. @vlachoudis did not merged my latest fixes yet, so there are still some bugs related to islands in upstream repository. Not sure what takes so long, perhaps @vlachoudis and @buschhardt are on vacation :-)

Harvie:
I see there's some controversy about licencing. is that reason why merging takes so long?

Just to make things clear:

This pull request introduces some code that is BSD and MIT licensed. Both of these are GPLv2 compatible. But i needed to add files indicating that the code was originaly BSD/MIT and was sublicensed as GPLv2. (You can use BSD and MIT code in GPLv2 project as long as you let everybody know about origin of that code and it's original license, which is what i've done here)

Harvie:
@vlachoudis i am concerned about my work as @MARIOBASZ works on the same code and the conflicts start appering. We need to synchronize in upstream to prevent this. I don't really care about licensing of my code as long as it will be free. We can figure this along the way, it's not that hard. I know we reuse some 3rd party code, but that can be solved also. Usualy opensource guys want their code to be as usefull as possible and various licenses are just tools to achieve that. It's not written in sone.

Can you please look at this? It's been almost month. I've broken some stuff in the past and this pull request fixes that, so i feel bad when it's left unmerged. If you have any concerns about the code quality, feel free to discuss it.

I used to be very enthusianistic about working on bCNC, but this heavily demotivates me. I am sorry to rush this, but i don't want to run in git conflicts and have to manualy rework my code. I know that you are not working on bCNC full time, so you have other stuff to do, but the cooperation between contributors is kinda hard if you don't pull regulary.

Once again sorry for rushing this and let me know if you need to help with something! :-)

vlachoudis:
Many thanks, there is a lot of work done here!

PR #913, opened 2018-07-09
Trochidal Icon

MARIOBASZ:
trochoidal


Harvie:
try this in your repository:

git pull https://github.com/vlachoudis/bCNC.git

MARIOBASZ:
thanks @Harvie, but I still do not know how to correct it

Harvie:
You put icon directly into the project root. That's probably not what you intended:

https://github.com/MARIOBASZ/bCNC/blob/master/trochoidal.gif

MARIOBASZ:
I can not delete the wrong file

MARIOBASZ:
very interesting! I do not know if you have corrected it, but if I move the silhouettes considered as islands, the "islands" continue in the same place, and I do not see how to eliminate them (previously you could eliminate the marks in tabs) Can you do something similar to unmark islands?

Harvie:
select and click "island" again. or just delete "island" from block name in editor.

Harvie:
This is not good. You've re-commited some of my changes to your fork and pull request, this will lead to conflict if this gets merged. Please don't do this!

MARIOBASZ:
ok, if I understand correctly: I upload to my branch the files as they are in the main branch so as not to generate conflicts! I assumed that after merging your changes, they would be compared or authorized with them by my aggregates, to avoid conflicts make a warning about what are after your work

PR #979, opened 2018-09-26
Update sketch.py

sfinexer:
1) Possibility of accidentally first point for the line.
2) The algorithm of search of continuations of lines is improved
3) Choice of the fading level
ORIGINAL:
line1

OLD programm:
screenshot_20180926_130426
NEW programm:
screenshot_20180926_130445
OLD programm:
screenshot_20180926_131032
NEW programm:
screenshot_20180926_131104

PR #981, opened 2018-09-26
Optimize function

sfinexer:
It is corrected.
In case the speed of CNC.feedmax_x CNC.feedmax_y different optimize correctly works!

### DIFF
--- a/CNC.py
+++ b/CNC.py
@@ -4350,6 +4350,8 @@ class GCode:
 				y2 = block.sy
 				dx = x1-x2
 				dy = y1-y2
+				dx/=CNC.feedmax_x
+				dy/=CNC.feedmax_y
 				matrix[i][j] = sqrt(dx*dx + dy*dy)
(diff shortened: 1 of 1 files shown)


Harvie:
Can you please explain how this works? Why it's needed to divide distance by feedrate when rearanging blocks?

Harvie:
Also this commit is already in #979 
I don't see no need to merge it separately

sfinexer:
@Harvie 
Instead of the distance between the points apply arrival time.
This is necessary if different speeds of movement of the machine along individual axes.

Harvie:
And what will happen when somebody will use bCNC just as g-code editor without ever connecting it to the machine?

sfinexer:
@Harvie It is necessary to take parameters not of the controller and the bCNC config.

sfinexer:
Feed max x
Feed max y
Feed max z

PR #985, opened 2018-09-29
all threads

sfinexer:


Harvie:
Are you sure this is good idea?

sfinexer:
@Harvie I have combined all the branches.
But in this case, the merger was suggested erroneously.
But there are no conflicts of branches.
I will continue to merge everything into a single branch, which is complex to wipe out for a new opportunity for bCNC.

MARIOBASZ:
to be more verbose, I suggest that the change introduced in cnc.pi comparing with 1e-7 introduced by @walterwissmann

be comparing with ESPV, even more if the precision of BCNC

Harvie:
I guess that this is no longer needed, since there are some issues in Mario's branch, so i can't merge it right away. This is the reason i've been skeptical about this. If you want your changes to be merged, please provide branch with your changes only like indicated in #979  Otherwise your merge will have to wait until all issues with Mario's branch are resolved.

Don't get me wrong. I am glad, that you've been maintaining this combined branch, so people had working version in meantime, while upstream was broken and blocked. But now i need your changes isolated to merge them, so we can get things moving.

Thanks

sfinexer:
Ok.

PR #1090, opened 2018-12-09
Update makefile, bcnc.pot and language-po and -mo files

SteveMoto:
The new directory "/ bCNC / controllers" was missing in the makefile file.
Directory was added and a new master translation file (bCNC.pot) was generated.
Numerous old translations could be reused.

The translations for the re-implemented plugin "midi2cnc" are also in the new bCNC.pot file.
The old translations of the plugin could be reused.

All * .po and * .mo files have been synchronized with the new master translation file (bCNC.pot - "POT-Creation-Date: 2018-12-09 20: 15 + 0100).


Harvie:
Thanks

PR #1092, opened 2018-12-09
Update language: de (po & mo)

SteveMoto:
"PO-Revision-Date: 2018-12-09 22:38+0100


treki1:
I edited/updated the german language file.
If you want, you can update these two new files (mo and po).
[locale_de.zip](https://github.com/vlachoudis/bCNC/files/7129742/locale_de.zip)

PR #1095, opened 2018-12-10
Code cleanup E301 E302 W391

SteveMoto:
E301 expected 1 blank line, found 0
E302 expected 2 blank lines, found 0 or 1
W391 blank line at end of file


Harvie:
What kind of software do you use to do this? AFAIK till now we were using single line after methods. I might not agree with that, but that's what everybody used. Is that some kind of requirement for python 3?

SteveMoto:
I used eric6 (IDE) and the integrated code checker.
The code checker is based on Python's PEP8 style guide -> https://www.python.org/dev/peps/pep-0008/ and on PEP 257 Docstring Conventions -> https://www.python.org/dev/peps/pep-0257/

bildschirmfoto-1

Harvie:
OK. I will look at it.
Just for future reference: I am not exactly fan of indenting using spaces, so i hope we'll keep tabs for now :-)

Harvie:
BTW i am also not really huge fan of doing automatic code fixes like this too often. I like the idea in general.
But in git it can cause troubles when merging code of multiple developers. If you use some script to change all files at once, it quite easily creates edit conflicts across all developer branches. (when compared to situation when each contributor edits different file and sometimes they edit the same file, but the changes are minor).

So in brief: Automatic code fixes are good, but we should not do them too often if they don't have effect on functionality. And we should avoid them altogether when there are huge pull requests pending (which is not the case today). I guess doing this stuff once a year is enough. Ideal solution is having git hooks to check/fix the code standarts at commit/push time, so we don't have to repair that in separate commits too often.

PR #1098, opened 2018-12-13
Update makefile

SteveMoto:
- Added new subdirectories for future translation work.
- pygettext.py removed -> xgettext works fine
- Added new subdirectories for removing * .pyc * .pyo files


Harvie:
Thanks

Issue #1102, opened 2018-12-17
Delay in tool change spindle start


leofaoro:
Hello
I use the tool change feature, and when I am done with putting the new tool in, the spindle starts and quickly moves to the workpiece without having spun to the target spindle speed. Would you please add a delay command of 5s or so so the spindle can speed up before plunging into the workpiece?
Thanks

Harvie:
Good point! Do you dare to create pull request? :-)

leofaoro:
While I know some python, I have never made a contribution. I'd love to learn this.

MARIOBASZ:
Look at the file cnc.py, it deals with the tool change in the sector:
code to change manually tool

sfinexer:
I think this is the prerogative of G-code and not the program B-cnc

leofaoro:
Thanks. I know where to find this. I have to learn the process of using GIT.

leofaoro:
> I think this is the prerogative of G-code and not the program B-cnc.
> 
> And if the tool is the filament?
> That at program delay its overheating is possible (for example in pyrography)?

hmm. this is true, but AFAIK bCNC is not for 3D printing, but I may be mistaken. I know using it with laser is also supported. I'll let the maintainers decide. I added a pull request.

sfinexer:
> hmm. this is true, but AFAIK bCNC is not for 3D printing, but I may be mistaken. I know using it with laser is also supported. I'll let the maintainers decide. I added a pull request.

B-cnc is used to send G-code, including for 3D models (engraving, burning, printing).

G-code should be sent without any changes.
When creating a G-code, its author (automatic script or otherwise) should take into account all the features of the machine in G-code.

Harvie:
I guess the pause can be optional selectable in UI...

![image](https://user-images.githubusercontent.com/276504/50297964-4dd97c80-047e-11e9-99fa-f27eb530fc87.png)

MARIOBASZ:
Thanks and congratulations for contributing a plugin. Maybe create a new field to choose waiting time before starting spindle. In the probe / tool change sector or in the naquina configuration.

leofaoro:
I modified my version to include an option to set the time for the spindle to wait. I am away from my CNC so I can't test. Can I submit as a pull request and so someone can test before merging? Newbie here so just learning how to collaborate. thanks

hamster65:
Hi,
[...]

PR #1104, opened 2018-12-19
add 5s pause after tool change

leofaoro:
to allow spindle to speed up

### DIFF
--- a/bCNC/CNC.py
+++ b/bCNC/CNC.py
@@ -1752,6 +1752,7 @@ class CNC:
 		lines.append("g0 x[_x] y[_y]")	# ... x,y position
 		lines.append("g0 z[_z]")	# ... z position
 		lines.append("f[feed] [spindle]")# ... feed and spindle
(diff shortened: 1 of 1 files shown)


sfinexer:
https://github.com/vlachoudis/bCNC/issues/1102

Harvie:
Thanks!
If you manage to add some UI to make this optional, it would be even better.
Isn't 5s too much? VFD spindle starts typicaly around 2 seconds, so 3 seconds should be enough.

BTW have you tried enabling pause after toolchange?

image

It might have similar result...

Harvie:
Also congratulations for your first commit merged to bCNC :-)

leofaoro:
> Also congratulations for your first commit merged to bCNC :-)
Thanks @Harvie ! I am learning python and how to contribute here. Hopefully many more to come!

I haven't tried this no proble manual tool change, it's not something I use. I really only use the WCS option.

Harvie:
I am not talking about "no probe", i am talking about "pause before and after probing"

PR #1116, opened 2018-12-23
python3

sfinexer:


Harvie:
This is not how it works. Also why you create 10 pull requests for the same thing?

Harvie:
You can only create one pull request and add more changes to it by commiting to the same branch.
However surely i will not accept this:

```
- #!/usr/bin/python
+ #!/usr/bin/python3
```

We plan to support both python2 and python3, so there's no reason to specify python version in files.
Also changing python hasbang to python3 will not magicaly make bCNC work in python3. There are much greater problems. Namely with unicode handling.

You can see v3 branch: https://github.com/vlachoudis/bCNC/tree/v3
There were some efforts of python3 porting. But it's a lot of work. There are even some automated scripts to change from python to python3, but it's still not enough and it needs lot of manual changes in whole bCNC codebase.

Harvie:
Also note that hashbang in plugins is not applied at all, because they are not executed. they are imported. Even bCNC should now be loaded as module rather than using hashbang. So there is no point of changing it.

sfinexer:
@Harvie, Why not just stop supporting Python 2 as outdated?

Harvie:
@sfinexer what do you mean? bCNC can't be run on python3 right now. If we stop supporting python2 there will be no single python version to run bCNC.

I want to fix some problems and release 0.9.15 (python2), next step would be to get bCNC working on python3 (bCNC v 0.9.33). Maybe there will be 0.9.16 or 0.9.17 in meantime if we find some more stuff that needs fixing https://github.com/vlachoudis/bCNC/milestones

I've spent several nights trying to get bCNC working on python3, even @vlachoudis, @coddingtonbear and maybe @SteveMoto were trying to get it working if i remember correctly. We agreed that it's not trivial task. But we also agree it's good idea to do it. We've managed to get bCNC partialy working on python3. Eg. main window opens and like 10% of features are working. But there are lots of problems and it's basically unusable. If you know how to port bCNC to work completely on python3 i would like to see it.

But we don't merge PARTIAL python3 ports to master. If you provide me with complete python3 port of latest bCNC version, i will be happy to merge it.

sfinexer:
@Harvie It seems to me on Python 3 I need to release version 1.0. And just do change the version numbering. The numbering is similar to that nowhere in B-cnc.

Harvie:
I don't care. Version number is just number. I care about code. We don't have working python3 code. That's what matters.

As of version 1.0 i think it should be something special in terms of usability and stability. Python3 is just rehashing of old stuff in slightly different syntax, so i don't think it's good idea to publish it as 1.0. Also python3 port will create lots of bugs and it will take lot of time and versions to get stable again. So i don't think it's good idea to release python3 version as 1.0 as it's going to be very buggy. Trust me.

That said i am not going to release 1.0.x after 0.9.x, i am going to do 0.10.x, 0.11.x, etc... I don't want to release 1.0 without OpenGL and proper 3D CAM support...

PR #1123, opened 2019-01-01
Update Sender.py

mihointech:


Harvie:
Thanks for your contribution. You inspired me to add colors for more GRBL states, but unfortunately i changed the color a bit to match following logic:

RED: something is wrong
YELLOW/ORANGE: we are connected and ready to work
GREEN: we are working

i hope you don't mind :-)

PR #1139, opened 2019-01-18
Recover from stuck sio_wait in Alarm and Error states

dshaded:
Hi!
This is a simple fix that solves sender freezes in situations when G38.2 or similar commands followed by %wait result in grbl Alarm state. In such cases bCNC keeps sio_wait==True until hard reset clears that flag. Soft reset does not help if grbl is configured to use homing cycle because it remains in Alarm state after soft reset and sio_wait remains true.

This definitely works with GRBL1 but I am not so sure about GRBL0 or smoothie.

### DIFF
--- a/bCNC/controllers/GRBL0.py
+++ b/bCNC/controllers/GRBL0.py
@@ -38,7 +38,7 @@ class Controller(_GenericGRBL):
 			# stop waiting and go on
 			#print "<<< WAIT=",wait,sline,pat.group(1),sum(cline)
 			#print ">>>", line
-			if self.master.sio_wait and not cline and pat.group(1) in ("Idle","Check", "Alarm", "error"):
+			if self.master.sio_wait and not cline and pat.group(1) not in ("Run", "Jog", "Hold"):
 				#print ">>>",line
 				self.master.sio_wait = False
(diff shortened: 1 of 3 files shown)


Harvie:
Wow. I love this kind of stuff! I haven't tried the code yet, but this kind of stuff is so hard to debug and it bugs me lately. I am very happy and thankfull, that you managed to find and adress such problem. I would love to know more about how this works. I will surely merge this, but first i want to know what exactly this means, because i don't know much about the sender code yet...

BTW wouldn't it make sense to invert that list? Eg. instead of `("Idle","Check", "Alarm", "error")` use something like `not ("Run","Jog", "Hold")`. I don't know. I didn't investigated yet. It's just idea. Don't know what the consequences would be...

dshaded:
Yes, this sounds reasonable. I updated this branch and tested it on my router. Works fine.

dshaded:
And about the works.
sio_wait is set to True by serialIO upon pulling "%wait" command from the queue. When sio_wait is active serialIO ignores queue completely waiting for grbl to get into "Idle" or "Check" state. That's how %wait is supposed to work. Hard reset is the only other way to drop sio_wait back to False.

Issue #1035, opened 2018-11-08
Stop during operation causes WCS to change


sfinexer:
Stop during operation.
And the Z coordinate is not displayed correctly.

Harvie:
Not enough informations

sfinexer:
[drawings-of-tigers-tiger-drawings-related-keywords-amp-suggestions-tiger-drawings(270мм X 210мм).cncR.zip](https://github.com/vlachoudis/bCNC/files/2560991/drawings-of-tigers-tiger-drawings-related-keywords-amp-suggestions-tiger-drawings.270.X.210.cncR.zip)
and press stop in the process of the machine.

Harvie:
1.) Do you use latest GRBL version?
2.) What WPOS coordinates does GRBL report when this happens?
3.) Do you use any offsets, WCS or similar stuff in your setup or gcode?
4.) What safe-z you have set?

Harvie:
WTF are these?

![image](https://user-images.githubusercontent.com/276504/48199349-1befd900-e35c-11e8-924c-2399fe46fd52.png)

sfinexer:
1) grbl_1.1g
2) I press the stop button at any time.
3) 

> [Controller]
grbl_0 = 10
grbl_1 = 25
grbl_2 = 0
grbl_3 = 0
grbl_4 = 0
grbl_5 = 0
grbl_6 = 0
grbl_10 = 3
grbl_11 = 0.0
grbl_12 = 0.002
grbl_13 = 0
grbl_20 = 1
grbl_21 = 0
grbl_22 = 1
grbl_23 = 3
grbl_24 = 100.0
grbl_25 = 500.0
grbl_26 = 25
grbl_27 = 20.0
grbl_30 = 50000.0
grbl_31 = 2000.0
grbl_32 = 0
grbl_100 = 80.0
grbl_101 = 80.0
grbl_102 = 40.0
grbl_110 = 10000.0
grbl_111 = 4500.0
grbl_112 = 10000.0
grbl_120 = 3000.0
grbl_121 = 200.0
grbl_122 = 1000.0
grbl_130 = 319.0
grbl_131 = 429.0
grbl_132 = 300.0

4),,,,

@Harvie CP1251

Harvie:
> CP1251

I don't care. still not valid g-code

```
[Блок №91301]   
G0 Y267.127 X46.646
G1Z-5
G0Z1
[Блок №91302]   
G0 Y267.059 X47.965
[...]

PR #1140, opened 2019-01-18
Try to restore TLO value on controller unlock and reset

dshaded:
Hi!
Here are some efforts to make "Manual tool change (TLO)" policy more usable. Grbl zeroes TLO on every reset and unlock event. And bCNC keep showing stored TLO value until manual $# is issued and real value gets parsed or until tool change cycle is executed.
This behavior require manual sending of tool change commands after every controller unlock or reset. Forgetting to run a tool change cycle leads to either a soft-limit error (best case but still annoying) or a tool crush.
I am not sure if these changes cover every possible TLO reset scenario. And the code is really ugly, especially use of purgeControllerExtra. But it makes my work with bCNC much easier. Not sure how to code it in a better way.

### DIFF
--- a/bCNC/ProbePage.py
+++ b/bCNC/ProbePage.py
@@ -338,7 +338,8 @@ class ProbeCommonFrame(CNCRibbon.PageFrame):
 	def loadConfig(self):
 		ProbeCommonFrame.fastProbeFeed.set(Utils.getFloat("Probe","fastfeed"))
 		ProbeCommonFrame.probeFeed.set(Utils.getFloat("Probe","feed"))
-		ProbeCommonFrame.tlo.set(      Utils.getFloat("Probe","tlo"))
+		CNC.vars["TLO"] = Utils.getFloat("Probe","tlo")
+		ProbeCommonFrame.tlo.set(CNC.vars["TLO"])
 		cmd = Utils.getStr("Probe","cmd")
 		for p in PROBE_CMD:
 			if p.split()[0] == cmd:
--- a/bCNC/Sender.py
+++ b/bCNC/Sender.py
@@ -117,6 +117,7 @@ class Sender:
 		self._stop	 = False	# Raise to stop current run
 		self._pause	 = False	# machine is on Hold
 		self._alarm	 = True		# Display alarm message if true
+		self._controllerReset = False # Controller reset was just detected and some G-code vars should be restored (e.g. TLO)
 		self._msg	 = None
 		self._sumcline	 = 0
 		self._lastFeed	 = 0
@@ -772,6 +773,10 @@ class Sender:
 				# so don't stop
 				if self._runLines != sys.maxsize:
 					self._stop = False
+					if self._controllerReset:
+						TLO = CNC.vars["TLO"]
+						self.sendGCode("G43.1Z%s"%(TLO))	# restore TLO
+						self._controllerReset = False
 
 			#print "tosend='%s'"%(repr(tosend)),"stack=",sline,
 			#	"sum=",sum(cline),"wait=",wait,"pause=",self._pause
--- a/bCNC/controllers/_GenericController.py
+++ b/bCNC/controllers/_GenericController.py
@@ -57,6 +57,7 @@ class _GenericController:
 		pass
 
 	def hardReset(self):
+		TLO = CNC.vars["TLO"]
 		self.master.busy()
 		if self.master.serial is not None:
 			self.hardResetPre()
@@ -67,24 +68,30 @@ class _GenericController:
 		self.master._alarm = False
 		CNC.vars["_OvChanged"] = True	# force a feed change if any
 		self.master.notBusy()
+		self.master.sendGCode("G43.1Z%s"%(TLO))	# restore TLO
 
 	#----------------------------------------------------------------------
 	def softReset(self, clearAlarm=True):
 		if self.master.serial:
 			self.master.serial.write(b"\030")
+		self.purgeControllerExtra()
 		self.master.stopProbe()
 		if clearAlarm: self.master._alarm = False
 		CNC.vars["_OvChanged"] = True	# force a feed change if any
 
 	#----------------------------------------------------------------------
 	def unlock(self, clearAlarm=True):
+		TLO = CNC.vars["TLO"]
 		if clearAlarm: self.master._alarm = False
 		self.master.sendGCode("$X")
+		self.master.sendGCode("G43.1Z%s"%(TLO))	# restore TLO
 
 	#----------------------------------------------------------------------
 	def home(self, event=None):
+		TLO = CNC.vars["TLO"]
 		self.master._alarm = False
 		self.master.sendGCode("$H")
+		self.master.sendGCode("G43.1Z%s"%(TLO))
 
 	def viewParameters(self):
 		self.master.sendGCode("$#")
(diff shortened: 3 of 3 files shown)


Harvie:
Don't you think this problem might be related to this issue? I've recently added GRBL reset/cleanup after each cycle. But i was reccomended to do so by GRBL developer.

https://github.com/vlachoudis/bCNC/issues/1035

dshaded:
Yes this seems related. But I'm sure that TLO policy was somewhat buggy and difficult to use even with earlier releases like 0.9.11.
And my first commit was not enough. Tool change macro still failed to properly set G43.1. That's how it happened:
- suppose grbl had just finished execution of tool change macro and changed state to Idle.
- serialIO parsed that state and called controllerStateChanged -> jobDone -> purgeController -> softReset -> purgeControllerExtra -> unlock
- ulock added TLO recovery code to queue and purgeControllerExtra added 1sec timeout to wait for grbl response.
- 1 second elapsed and serialIO resumed working. And the first thing it received from the port was "Grbl 1.1f" hello. Parse line handled that and set the _stop flag.
- serialIO started the next work cycle, saw the _stop flag and called purgeQueue loosing TLO recovery code.

So I commited another crutch to fix this. And today it really worked as intended. Although I had to delete z-axis return code from my main branch to get rid of irritating soft-limit errors.

This patch is still ugly. Maybe I will find some time to make it better. Actually I'd like to cleanup sender code, decouple it from GUI Application class and add some tests but it looks like a lot of time.

Harvie:
Sorry this is taking so long to merge, but i would like to hear some feedback from people who are using TLO. I've never used it, so i don't have opinion on whether this is good idea...

dshaded:
I have looked through related issues here and googled for TLO usage with grbl. It looks like people never use this option because it is too hard to. Most mentions are like "do not use tlo cause it is buggy and works only in big machines world, just choose wcs here".
And the only real benefit of TLO is the ability to use different WCS within one program which is rarely needed for smaller machines. As far as i know.
So I am not sure if we will receive any feedback on this patch before it gets merged.

sirusdv:
Funny enough when I probe I set the tlo button to change wcs:

```
diff --git a/ProbePage.py b/ProbePage.py
index 49425be..4db58fb 100644
--- a/ProbePage.py
+++ b/ProbePage.py
@@ -269,7 +269,7 @@ class ProbeCommonFrame(CNCRibbon.PageFrame):
 	def tloSet(self, event=None):
 		try:
 			CNC.vars["TLO"] = float(ProbeCommonFrame.tlo.get())
-			cmd = "G43.1Z"+str(ProbeCommonFrame.tlo.get())
+			cmd = "G92Z"+str(ProbeCommonFrame.tlo.get())
 			self.sendGCode(cmd)
 		except:
 			pass
@@ -308,6 +308,7 @@ class ProbeCommonFrame(CNCRibbon.PageFrame):
 		ProbeCommonFrame.fastProbeFeed.set(Utils.getFloat("Probe","fastfeed"))
 		ProbeCommonFrame.probeFeed.set(Utils.getFloat("Probe","feed"))
 		ProbeCommonFrame.tlo.set(      Utils.getFloat("Probe","tlo"))
+		CNC.vars['TLO'] = Utils.getFloat("Probe","tlo")
 		cmd = Utils.getStr("Probe","cmd")
 		for p in PROBE_CMD:
 			if p.split()[0] == cmd: 
```

that `CNC.vars['TLO'] = Utils.getFloat("Probe","tlo")` probably should be added upstream.

Harvie:
@sirusdv why you need that patch? You can just set bCNC to use WCS instead of TLO:

image

Harvie:
In my lack of understanding of this problem, i've resorted to ask GRBL developers to reccomend us how to handle this situation: https://github.com/gnea/grbl/issues/623

dshaded:
G92 is not the same as setting WCS tool change mode in bCNC. It is a global offset added to all WCS. Just like TLO but also with X and Y support. Unfortunately grbl docs state that it is also not persisted between controller reboots.

And asking grbl devs is really a good idea. I'll watch that issue.

Harvie:
> G92 is not the same as setting WCS tool change mode

Can you then add G92 tool change mode to bCNC so that we can choose it from menu as another option?

dshaded:
It should be easy but what for? G43.1 and G92 are acting identically regarding Z-axis offset and G43.1 is specifically designed to set tool length offset. I don't see any benefits but more confusion.

Harvie:
> I don't see any benefits but more confusion.

Ok. I was thinking that there are some benefits, since you've patched the bCNC to use this.

Harvie:
According to discussion with GRBL developers i think i've finaly made my mind. I think that persistent TLO is a good thing. But there is some stuff that i really want to handle differently than it's currently implemented in this PR.

1.) bCNC should query GRBL using `$#` command every 5 seconds or so (similar to `?` query, but less frequent).
2.) whenever bCNC receives parameter data, eg.: `[TLO:0.000]` it should be stored to CNC.vars.
3.) it should be restored on GRBL reset

TLO should never be updated in CNC.vars directly. It should be set by sending gcode to GRBL and then querying the GRBL to make sure it has applied. The TLO settig gcode should be immediately followed by call to `viewParameters()` (bCNC's abstraction to send `$#` to GRBL or similar command to other types of motion controllers). That way TLO will get immediately send from GRBL back to bCNC to be stored in CNC.vars.

TLO should not be stored in bCNC.ini or loaded after bCNC restart as it might cause it to be accidentaly stored on one CNC machine and then loaded to different one if you are using bCNC with multiple machines.

We might want to figure out if it would make sense to apply the same behaviour for the following GRBL $# parameters as we can query/store these along with TLO using the same code:
```
 [G28:0.000,0.000,0.000]
 [G30:0.000,0.000,0.000]
 [G92:0.000,0.000,0.000]
```

**So if you rework the PR to reflect this concept, i will merge it.**

dshaded:
I'll think about this changes but first I'd like to try to patch grbl. It's a bit complicated considering how persistense of G29/G30 is implemented and how tight are flash size limits. But still worth a try.

Harvie:
> I'd like to try to patch grbl

That makes sense. Thanks.

Harvie:
I've done some modifications, so that GRBL parameters are stored to CNC.vars upon receiving...
Also i've added GUI for TLO and G92. Everytime you change TLO you just need to call `app.mcontrol.viewParameters()`. This will poll GRBL for TLO and update the GUI automaticaly.

Problem is that we can't poll GRBL for TLO unless it's in idle state. So i can't update it during job. At least i don't know how.

Harvie:
I've also noticed there is already following line in method `purgeController()` in file `controllers/_GenericController.py` :

```python
                self.master.sendGCode("G43.1Z%s"%(TLO)) # restore TLO
```

dshaded:
Changes to GUI are definitely ok. And quering/parsing of $# is reasonable too.
It will probably be enough to send G43.1 and G92 back to Grbl when controller reset is detected to prevent their loss. But I still would like to wait for a reply to my pull request https://github.com/gnea/grbl/pull/624 .

Speaking of storing these params to bCNC.ini. I don't think that it will break anything because bCNC already keeps machine-specific calibration data there. Running recalibration on every program restart doesn't sound like a good idea.

As for the code in purgeController() you can check back at my [2nd message](https://github.com/vlachoudis/bCNC/pull/1140#issuecomment-456047642) in this thread. This code doesn't work for the very same reason - in some cases it adds restore command to a queue from the sender thread. And the queue gets purged when Grbl hello is detected and before G43.1 gets executed.

dshaded:
It's been more than a year now. I've been using modified grbl with [this tiny patch](https://github.com/dshaded/grbl/commit/6994d617253d93e31a4b0049d26484a845990b8b) all the time and it works like a charm. Not a single case of unexpected TLO loss.
Patching grbl manually is really simple and I don't think that tracking TLO in bCNC is better solution.

PR #1149, opened 2019-01-29
Trochoid Path - Helical descent - Scale

MARIOBASZ:
Small corrections to helical descent
Corrected generation of adaptive path for trochoidal cutting: Should work.
Caution: Not yet implemented in trochoidal cutting.
Note: Harvie, the start and end of the adaptive path are indicated by two movements: a small movement (0.00001mm) and a retraction of the same magnitude. It would be good to change it to a brand that indicates it, so as not to generate inconveniences if you want to move the generated route due to rounding problems.
I can share with you how I solved it in my trocoidal_3d, the idea is simple: To the start alarm I measure the next segment. At distance 0, trochoid radius = trochoid radius. Distance = measured distance, (at the end), trochoid radius = 0 (you can then establish if there is a minimum adaptive trocoid diameter, my idea for solution was another).
When returning the path, it also does adaptive, does not generate the trochoids.

Scale multiple shapes. Scale from the center of the shape (only for the individual selection of shapes)


Harvie:
See the travis CI. there is some failing check. click on details here on github. Then go to job for "python 2.7 on linux" and you will see what is the problem. for example here you can see it for your current commit:
https://travis-ci.com/vlachoudis/bCNC/jobs/174237283

MARIOBASZ:
Today I saw that there is some conflict with climbing, tomorrow I will review and try to correct it (perhaps the icon name only?)

MARIOBASZ:
Corrected errors in the scale's plugin.
Observation: You can choose several silhouettes, but with limitations:
1-Works correctly without the "X Y centered" option
2- If "X Y centered" is chosen, all the silhouettes are scaled from the center of the first figure
3-Center works correctly for regular figures and more than three sides (in the triangle generates the center of functions in the wrong place, it is a problem of the center function)
4- For multiple figures, I could not generate the result with the corresponding name of the block (ask for help to improve this)
Harvie, where I find the "double click" help for the plugin.
Hugs

Harvie:
>  Harvie, where I find the "double click" help for the plugin.

```
		self.variables = [
			("adaptative",  "bool",                   1,   _("Adaptative"), "PUT IT HERE"),
```

MARIOBASZ:
Thanks

PR #1157, opened 2019-02-07
Update text.py - Preparated to be separated into blocks

MARIOBASZ:
Text prepared to be separated into blocks.
Only "cut here" was added at the end of each contour.
Executing Split divides the text one block per in each closed contour. Execute Join joins in a single block.


MARIOBASZ:
Example
image
[example.text.zip](https://github.com/vlachoudis/bCNC/files/2842882/example.text.zip)

Issue #1108, opened 2018-12-22
B-cnc stops sending data


sfinexer:
As can be seen in the video, data is accumulated in the Arduino buffer, although in response to all the data the controller sends the correct response.
Stop occurs in a random place G-code.

http://cnc.tiven.ru/ssr-2018-12-22_18.27.31.mp4

Code arduino:
```c
void setup() {
    Serial.begin(115200);
}
void loop() {}

String inputString = "";  
void serialEvent() 
{
    while (Serial.available()) {
    char inChar = (char)Serial.read();  // get the new byte
    inputString += inChar;
    if (inChar == '\n') {
      //Serial.println(inputString);
      Serial.println("ok");
      inputString = "";
      delay(5);
      }
    }
}

```

Harvie:
> I got the impression that the "ok" commands are going too fast, and B-cnc does not have time to accept them, or process them.

Add delay to arduino code.

Does this happen with real GRBL? If not, this issue would be irrelevant.

sfinexer:
@Harvie Not. There are many variations of offline grbl on the internet. But grbl itself is not provided for this mode of operation.
There is an idea to remake grbl so that there was an opportunity to work with SD-card. For this, according to the specification, grbl, upon receipt of a command, confirms it with "ok", thereby having to emulate data acquisition.
I implemented it, but the data ceases to be sent for reasons unknown to me.
And in this way, B-cnc will be able to transfer all code at once into a buffer, which will be executed independently of B-cnc.

Harvie:
You just suppose that your GRBL protocol implementation is correct. But i think this might not be the case. I think that GRBL is not supposed to respond to `?` command with `ok`. And there can be even more similar cases that we are not aware of.

Harvie:
I've just tried to run big job on my machine and it stoped sending after ~45k lines of g-code.
Now trying UGS platform sender, but so far so good...
I guess we have to figure this out...

sfinexer:
> I've just tried to run big job on my machine and it stoped sending after ~45k lines of g-code.
> Now trying UGS platform sender, but so far so good...
> I guess we have to figure this out...

I have not yet verified the command using spy.
But be sure to do it.

Harvie:
https://github.com/grbl/grbl/wiki/Interfacing-with-Grbl#streaming-a-g-code-program-to-grbl
https://github.com/grbl/grbl/blob/master/doc/script/simple_stream.py
[...]

PR #1166, opened 2019-02-18
Update fake-grbl-arduino.ino

sfinexer:
Having looked that gives out GRBL in the code lacks the response "OK".

### DIFF
--- a/tests/fake-grbl-arduino/fake-grbl-arduino.ino
+++ b/tests/fake-grbl-arduino/fake-grbl-arduino.ino
@@ -21,12 +21,10 @@ void serialEvent()
             buff[readcnt] = 0;
             if ((strncmp(buff, "?", 1) == 0)) {
                 printPgmString("<Idle|MPos:0.000,0.000,0.000|FS:0,0|WCO:0.000,0.000,0.000>\n");
-                Serial.println("ok");
                 break;
             }
             if ((strncmp(buff, "$G", 2) == 0)) {
                 printPgmString("[GC:G0 G54 G17 G21 G90 G94 M5 M9 T0 F0 S0]\n");
-                Serial.println("ok");
                 break;
             }
             printPgmString("ok\r\n");
--- a/tests/fake-grbl-arduino/fake-grbl-arduino.ino
+++ b/tests/fake-grbl-arduino/fake-grbl-arduino.ino
@@ -20,11 +20,13 @@ void serialEvent()
         {
             buff[readcnt] = 0;
             if ((strncmp(buff, "?", 1) == 0)) {
-                printPgmString("<Idle|MPos:0.000,0.000,0.000|FS:0,0|WCO:0.000,0.000,0.000>\r\n");
+                printPgmString("<Idle|MPos:0.000,0.000,0.000|FS:0,0|WCO:0.000,0.000,0.000>\n");
+                Serial.println("ok");
                 break;
             }
             if ((strncmp(buff, "$G", 2) == 0)) {
-                printPgmString("[GC:G0 G54 G17 G21 G90 G94 M5 M9 T0 F0 S0]\r\n");
+                printPgmString("[GC:G0 G54 G17 G21 G90 G94 M5 M9 T0 F0 S0]\n");
+                Serial.println("ok");
                 break;
             }
             printPgmString("ok\r\n");
--- a/tests/fake-grbl-arduino/fake-grbl-arduino.ino
+++ b/tests/fake-grbl-arduino/fake-grbl-arduino.ino
@@ -20,15 +20,14 @@ void serialEvent()
         {
             buff[readcnt] = 0;
             if ((strncmp(buff, "?", 1) == 0)) {
-                printPgmString("<Idle|MPos:0.000,0.000,0.000|FS:0,0|WCO:0.000,0.000,0.000>\n");
+                printPgmString("<Idle|MPos:0.000,0.000,0.000|FS:0,0|WCO:0.000,0.000,0.000>");
                 break;
             }
             if ((strncmp(buff, "$G", 2) == 0)) {
-                printPgmString("[GC:G0 G54 G17 G21 G90 G94 M5 M9 T0 F0 S0]\n");
+                printPgmString("[GC:G0 G54 G17 G21 G90 G94 M5 M9 T0 F0 S0]");
                 break;
             }
             printPgmString("ok\r\n");
         }
-        //delay(5);
     }
 }
--- a/tests/fake-grbl-arduino/fake-grbl-arduino.ino
+++ b/tests/fake-grbl-arduino/fake-grbl-arduino.ino
@@ -18,6 +18,7 @@ void serialEvent()
         readcnt = Serial.readBytesUntil(0x0A, buff, sizeof(buff));
         if (readcnt > 0) 
         {
+            buff[readcnt] = 0;
             if ((strncmp(buff, "?", 1) == 0)) {
                 printPgmString("<Idle|MPos:0.000,0.000,0.000|FS:0,0|WCO:0.000,0.000,0.000>\r\n");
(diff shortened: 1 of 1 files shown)


Harvie:
I don't think that GRBL sends `OK` in response to `?` command

sfinexer:
@Harvie Video:
http://cnc.tiven.ru/ssr-2019-02-19_08.13.52.mp4

Harvie:
See this log made using pyserial `spy://`. I don't see any OK reply for `?` command sent by GRBL:

```
?<Idle|MPos:0.000,0.000,0.000|FS:0,0|Ov:100,100,100>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0|WCO:-3.236,0.430,-1.587>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0|Ov:100,100,100>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0|WCO:-3.236,0.430,-1.587>
?<Idle|MPos:0.000,0.000,0.000|FS:0,0|Ov:100,100,100>
```

Harvie:
The reason why you get `ok` from grbl is because you don't send `?` the arduino ide actually sends `?\n`. Know the difference.

sfinexer:
Then send a fake-grbl-arduino.ino long G-cod and verify that the buffer is full.

Harvie:
Because you are sending `?` using terminal, which also sends `\n`. But during normal operation it does not send `ok`.

sfinexer:
@Harvie Probably you are right.
But if you do not send "OK" then after a while the programs stop.
https://github.com/vlachoudis/bCNC/issues/1108

Harvie:
> But if you do not send "OK" then after a while the programs stop.

This might be a bug in bCNC, but i was not able to replicate it... Surely you will not fix it by arduino sending more `ok`s than stock GRBL. It will only hide the bug if there's any. Also it's possible that you have RF interference with serial line that prevents `ok` from being delivered...

sfinexer:
The text "<Idle|MPos:0.000,0.000,0.000|FS:0,0|WCO:0.000,0.000,0.000>\r\n" is too long, having studied the GRBL code, the function printPgmString is used there, I copied it here.
It is necessary to test better, but the hang of the program at the moment is not observed.

sfinexer:
I'm unsure of the correctness of changes, I ask users to test these changes.

sfinexer:
I connected the Arduino with a short wire yesterday. And without "OK" she stopped ((((

Harvie:
> without "OK" she stopped

Obviously. That's how GRBL streaming works.
https://github.com/gnea/grbl/wiki/Grbl-v1.1-Interface#streaming-a-g-code-program-to-grbl

sfinexer:
@Harvie. Thank you very much.

So the patch is correct in terms of documentation.
But the question of stopping remains open.
The documentation for the Arduin, I think is not entirely correct, it says about the buffer size of 128 characters. However, UNO uses a buffer of 80 characters (grbl file config.h "#define LINE_BUFFER_SIZE 80 // Uncomment to override default in protocol.h").
But in my example, we use fake-grbl-arduino.ino and there in the code of my patch char buff [128];
However, the behavior of the program remains unpredictable.

sfinexer:
@Harvie 
https://github.com/gnea/grbl/wiki/Grbl-v1.1-Interface#real-time-status-reports
It says that GRBL in theory may not respond to a state request (?) At all.
I have a question, does B-cnc expect an answer?
Judging from the description of the GRBL, the answer is optional, and B-cnc is not obliged to expect a response.
Indeed, in my case, there are no delays in performing the state response (?), Since fake-grbl-arduino.ino constantly "performs" the G-cod, and it does not have time to respond to requests for the state.

Harvie:
> I have a question, does b-CNC expect an answer?

i think it does not expect answer. except when the `?` is issued manualy by user in terminal. In that case user will receive one status report printed to terminal. Other status reports are then silently procesed in background. Otherwise there would be hundreds of status reports printed in terminal all the time.

sfinexer:
Accidentally deleted the patch.
I think it's worth the wait to read the SD card.
Probably, the problem is in the transmission of large data, since if the G-code contains less than 20k lines, the error is very rare.
I use G-code where about 150-200 thousand lines.

Harvie:
I think we can write arduino code for "serial loopback" and test the arduino, if the usb-serial chip works properly at 115200 bauds.

sfinexer:
In other words, the error is probably not in the response to the "?" or another command, namely the loss of data.

Harvie:
I don't know, but we should check this. I few months agot i had nasty problems with chinese usb-serial chips running at 115200, so i think we need some tools to test it. Something like `Serial.write(Serial.read());` Then send lots of data to it, let it run for 10 minutes and check md5 sum of the data or something. Also do the same tests with motors running to test for EMI problems.

sfinexer:
@Harvie 
Well, as I wrote earlier, I have a data buffer on the Arduino2560 (There are nominally more RX and TX). But as I wrote earlier, to my disregard, the buffer is writing to the SD card, but I still have nothing to read it.

sfinexer:
Here is the test code for the Arduino 2560
[becap.ino.zip](https://github.com/vlachoudis/bCNC/files/2945813/becap.ino.zip)

sfinexer:
The system I use:
```
B-cnc ---> Arduino2560 ---> ArduinoUNO
                \            /
                   SD card
```

Harvie:
```C
void setup() {
  Serial.begin(115200);
}

void loop() {
  if (Serial.available()) {
    Serial.write(Serial.read());
  }
}
```

```bash
dd if=/dev/urandom of=random.bin bs=1M count=1
stty -F /dev/ttyUSB3 115200 cs8 ixoff
tail -f /dev/ttyUSB3 | pv > random2.bin

#Then in another terminal:
cat random.bin | pv > /dev/ttyUSB3
sleep 2 #wait until everything received
md5sum random.bin random2.bin

```

sfinexer:
@Harvie I will try.
But the question arises, what to do is md5sum not match?

sfinexer:
sfinexer@sfinexer:~/test$ tail -f /dev/ttyUSB1 | pv > random2.bin
00 B 0:00:23 [0,00 B/s] [<=>                                                                        ]

Harvie:
Pay attention to this line: #Then in another terminal:

sfinexer:
> Pay attention to this line: #Then in another terminal:

Thanks, I thought it after doing)

sfinexer:
@Harvie random2.bin is not increasing.

Harvie:
@sfinexer I've written better test for you:
https://github.com/vlachoudis/bCNC/tree/master/tests/arduino-loopback

Issue #1163, opened 2019-02-13
box tool dz bug


spanner888:
Looks like a copy paste error in the box tool.
		if nx > 0:
			self.nx = int(nx)
		else:
			self.nx = int(round(self.dx/nx))

		if ny > 0:
			self.ny = int(ny)
		else:
			self.ny = int(round(self.dy/ny))

		if ny > 0:
			self.nz = int(nz)
... so the code has TWO if ny > 0: .. and no if nz > 0:

Should the last one on line 74 actually be "if nz > 0:" ??

Harvie:
Does this cause any problems?

sfinexer:
Yes this is a mistake.
Of course, if nz> 0:

Harvie:
But can you explain what unexpected behaviour this causes? I don't want to read the code just to figure out what the real world consequence is, since you already know it.

Harvie:
Should be fixed by #1170

PR #1170, opened 2019-02-21
Update box.py

sfinexer:
1) Fixed division by 0 in case of zero number of teeth.
2) Removed unnecessary code as the parity check follows.
https://github.com/vlachoudis/bCNC/issues/1163


Harvie:
I guess it would make more sense to automaticaly calculate finger count based on desired width of finger and size of box, so all fingers uniform... Also we can specify the finger width as N times the material thickness. Eg. all fingers might be 4 times the board thickness for easy box making...

Issue #1169, opened 2019-02-21
Autolevel does not use rapid moves


FrenkK:
In 0.9.14, when I do grid levelling, all the moves are performed with the same low probe speed. This makes the levelling take longer than it needs to.
The program should use G0 Xx Yy Fsss for x-y direction moves after each probe (and possibly for lifting the probe too?). As speed value, the "Fast Probe Feed" value in the probing menu should be used. This would save a lot of time and would be more intutive.
A big thank you to everybody working on this awesome project! :)

Harvie:
Can you please provide pull request for this one? Sounds easy to fix...

FrenkK:
I'm a newbie at git and Python, but I'll try.

Harvie:
> I'm a newbie at git and Python

I also was when i first started contributing to bCNC few months ago :-) We need more developers :)

FrenkK:
Ok, this took way too much time, but at least it seems to work. I added the pull request and I now even have a basic idea how Github works :)

PR #1171, opened 2019-02-21
Fix for issue #1169

FrenkK:
While autolevel scanning, all the moves except probing now use the Fast Probe Feed speed setting. This speeds up autolevel scanning considerably. #1169 
Edit: Tested with Smoothieboard with latest CNC firmware (GRBL mode is default here). In my case, a 2 hour scan now takes 30 minutes.

### DIFF
--- a/bCNC/CNC.py
+++ b/bCNC/CNC.py
@@ -298,12 +298,12 @@ class Probe:
 		self.makeMatrix()
 		x = self.xmin
 		xstep = self._xstep
-		lines = ["G0Z%.4f"%(CNC.vars["safe"]),
+		lines = ["G0Z%.4fF%g"%(CNC.vars["safe"],CNC.vars["fastprbfeed"]),
 			 "G0X%.4fY%.4f"%(self.xmin, self.ymin)]
 		for j in range(self.yn):
 			y = self.ymin + self._ystep*j
 			for i in range(self.xn):
-				lines.append("G0Z%.4f"%(self.zmax))
+				lines.append("G0Z%.4fF%g"%(self.zmax,CNC.vars["fastprbfeed"]))
 				lines.append("G0X%.4fY%.4f"%(x,y))
 				lines.append("%wait")	# added for smoothie
 				lines.append("%sZ%.4fF%g"%(CNC.vars["prbcmd"], self.zmin, CNC.vars["prbfeed"]))
@@ -311,7 +311,7 @@ class Probe:
 				x += xstep
 			x -= xstep
 			xstep = -xstep
-		lines.append("G0Z%.4f"%(self.zmax))
+		lines.append("G0Z%.4fF%g"%(self.zmax,CNC.vars["fastprbfeed"]))
 		lines.append("G0X%.4fY%.4f"%(self.xmin,self.ymin))
 		return lines
 


Harvie:
I don't think you are supposed to set feedrate to G0 command, since G0 is already rapid... (machine moving as fast as possible)

FrenkK:
At least with Smoothieboard that is not the case. G0 just has a separate feedrate from G1 and you set it the same way.

Harvie:
According to smoothieware wiki http://smoothieware.org/g0#default-feedrate

> Default feedrate
> 
> If you have just started your Smoothieboard and have never sent a F parameter, your G0 moves will move at the default “seek” feedrate.
> 
> This is set in the configuration file by setting the default_seek_rate config option.

So i guess it should be enough if you set this to higher value?

Haven't checked yet, but i am afraid, that GRBL will not like `G0 F...` syntax.

FrenkK:
The simple fact is that the G0 feedrate does not remain set with smoothieboard. I tried literally everything, but I could not get it to use fast moves during scanning without this update.

Harvie:
even if you change default_seek_rate ?

FrenkK:
default_seek_rate gets lost as soon as I do a probe. Smoothieboard can do many things, but not always exactly like you would expect.
Anyway: I checked [here](https://wiki.shapeoko.com/index.php/G-Code#Motion_.28G.29) and the feedrate command (F) is treated separate (you can even put it as a separate line or in line with a move code). So, at least according to spec, this fix should have no particular effect on GRBL - fast G0 moves will use the fast feed anyway and the probing has the low probing speed specified in each line.

Harvie:
GRBL states that it follows LinuxCNC syntax, but there's no mention about ability to specify feedrate for G0:
http://www.linuxcnc.org/docs/html/gcode/g-code.html#gcode:g0
OTOH I've just tested and GRBL 1.1f ignores feedrate specified at G0

This is probably good, because it will not affect GRBL users.

FrenkK:
I have since moved my mill to GRBL due to other problems (jobs would not finish correctly in bCNC, it's impossible to stop a job without losing position etc.). But the fix in this pull request did work for Smoothieboard, solving the problem. If somebody want's to use it, it might be useful to put the hack in a Smoothie-specific part of code.

Harvie:
> But the fix in this pull request did work for Smoothieboard, solving the problem.

So this only fixes the probing, but there are still problems during cutting right?

> due to other problems I had with Smoothieboard

Do you know what causes these other problems? Luckily @vlachoudis said that he's working on fixing the smoothieboard support in bCNC. So, i'll leave this open, so he can review/merge it as he has experience with both GRBL and Smoothie.

FrenkK:
> So this only fixes the probing, but there are still problems during cutting right?
> Do you know what causes these other problems?
I think the problem where the run will sometimes not end correctly (program ends, but bCNC still acts like it is running - grey buttons etc.) is in fact a bCNC issue. It is already in the issues, I think.

The main cause for my switch is the fact that with Smoothieboard you can't stop program execution without either letting the buffer run empty (meaning CNC makes a few more moves before stopping, but bCNC does not support such a stop anyway) or losing position by resetting the board (ctrl-X, stop in bCNC). AFAIK, this is not something that will be changed soon. I'm an imperfect human being and sometimes I forget something at the start - it is really handy to have the option to just stop, correct the problem and continue without realigning everything from the start.

Harvie:
> with Smoothieboard you can't stop program execution without either letting the buffer run empty or losing position by resetting the board

Really? Is that smoothie problem or bCNC problem? GRBL is bit similar. You have to first issue the feed hold and once motors decelerate, you can safely reset GRBL to clean buffer without loosing WCS. Does smoothie have feed hold feature?

Harvie:
I can consider merging this, if we can be sure it does not interferre with non-smoothie hardware

FrenkK:
I think it should not interfere, all G1 moves should have speed specified anyway. But there is no guarantee for all possible use cases, so it's probably best to leave this one be and close it.

Harvie:
We can check if bCNC is configured to use smoothieboard or not...

PR #1181, opened 2019-03-05
Several fixes for Trochoidal Tool

jmomadit:
- Lead into and out of trochoidal path segment using small half circle for smooth motion
- Removed several unneeded linear moves which caused even backward motion because of rounding issues
- split circles into half circles, otherwise GBRL controller seems to have problems sometimes
- Comments added, some restructuring to improve readability
Issues were observed while milling an e-guitar body contour consisting of lines and arc segments,
fixes above solved them.
All in all a very usable great tool, thanks! :-)
Rgds, Juergen

### DIFF
--- a/bCNC/plugins/trochoidal.py
+++ b/bCNC/plugins/trochoidal.py
@@ -62,11 +62,13 @@ class Tool(Plugin):
 
 		#print("go!")
 		blocks  = []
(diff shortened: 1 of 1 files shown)


Harvie:
Before:

image

After:

image

Harvie:
Thanks.

PR #1202, opened 2019-03-25
Scaling Laser Option

MARIOBASZ:
Added an option for the laser. I do not have a laser, please confirm if it works correctly or there is an error.
Thanks Mario


bCNC/ToolsPage.py
```diff
@@ -654,6 +658,10 @@ def update(self):
 			self.master.cnc()["cutfeed"]  = self.fromMm("feed")
 			self.master.cnc()["cutfeedz"] = self.fromMm("feedz")
 			self.master.cnc()["stepz"]    = self.fromMm("stepz")
+			self.master.cnc()["zretract"] = self.fromMm("zretract")
+			self.master.cnc()["feedbeforecontact"] = self["feedbeforecontact"]
+			self.master.cnc()["hardcrust"] = self.fromMm("hardcrust")
+			self.master.cnc()["hardcrustfeed"] = self["hardcrustfeed"]
```
Harvie: I've already stated that i don't like this idea.

bCNC/lib/bpath.py
```diff
@@ -1474,14 +1474,75 @@ def trochovercut(self, offset, overcut, adaptative, adaptedRadius):
...
+					#only overcut
 					else:
 						self.insert(i,Segment(Segment.LINE, segment.A, segment.A + D))
 						self.insert(i+1, Segment(Segment.LINE, segment.A+D, segment.A))
 						i += 2
 			prev = segment
 			Op = prev.orthogonalEnd()
 			i += 1
+	#----------------------------------------------------------------------
+	def two_bit_adaptative_cut(self, offset, overcut, adaptative, adaptedRadius):
```
Harvie: What is this and why it should go to bpath.py ?

MARIOBASZ:
@Harvie, I have some things I work on my desk, I wanted to upload only the scale update, other things that I thought were deleted before uploading to the master, such as the modifications in tools page and in bpath, have gone up
What I find useful (saving time, softer entry) for me is understandable that it is not useful for another or results in a complication, but in any case it will be in another conversation if my point of view is advantageous.
Only scaling, and, I do not know if it is solved correctly for laser.
Two bit is something in development (roughing and finishing), not yet mature

PR #1221, opened 2019-04-27
Small fix once bCNC is installed

pci06:
Hello,
I've just add some try/except in order to avoid "startup" missing error.

regards.


Harvie:
Why don't you have these sections in your bCNC.ini?

Also i don't think it's good practice to use "except: pass", you should specify the type of exception that you want to ignore (usualy can get the name right from error message)...

Harvie:
Any news on this?

Harvie:
Realized it's already used on the other lines, so i guess it's not going to get any worse by merging this. Thanks for your effort!

Issue #1121, opened 2018-12-29
Pendant Web Interface - Question


qvisionsa:
I'm writing a new web interface for bCNC Pendant, using bootstrap 4 and font awesome.
When testing the interface without using bCNC web server, all looks fine.
Using bCNC web server, I miss all icon from font awesome. Is there some configuration parameter that I have to change or add?

Harvie:
Please coordinate your efforts with @PCSDIAS at #930

Harvie:
> I miss all icons from font awesome

Please chceck if the icons are served with proper mime-type HTTP header.

qvisionsa:
I wish you a happy new year!
Of course! I’m currently designing the interface and I can send you some screenshots and hear from you your thoughts.

Regards

Stelios Malakoudis
QVISION SA

________________________________
Από: PCSDIAS <notifications@github.com>
Στάλθηκε: Δευτέρα, Δεκεμβρίου 31, 2018 22:27
Προς: vlachoudis/bCNC
Κοιν.: QVision SA Mail; Author
Θέμα: Re: [vlachoudis/bCNC] Pendant Web Interface - Question (#1121)


Would it be possible to share with us the html, js and css files?

qvisionsa:
This is my first approach for machine control. As you can see, I plan to implement the interface using tabs. The screenshots are for an iphone 6.
I would like to hear your suggestions.
![screen2](https://user-images.githubusercontent.com/23384937/50569561-73b12c00-0d70-11e9-8033-15e6695e213d.JPG)
![screen1](https://user-images.githubusercontent.com/23384937/50569562-73b12c00-0d70-11e9-83ad-cdf4c9e8d64d.JPG)

PCSDIAS:
There is not much to invent in a bCNC GUI. IMHO the original layout has only two  problems:  1. resizing does not work very well  in the cell phone screen and 2. awkward jogging control.
How do you intend to address these issues in your project?

qvisionsa:
The first problem I will try to solve (and I'm close enough) is resize screen for small devices.
The second problem is that the screen space in small devices are not enough for all controls. My approach to use tabs and switching between them with a fingertip, give us the way to have more controls on the web pendant.
As soon as I finish coding, I will let you know.

PCSDIAS:
Dear Stelios Malakoudis



I have shared with you on
https://www.dropbox.com/s/giic9he6c2vjci1/bCNCPendant.zip?dl=0 an almost
[...]

PR #1230, opened 2019-05-26
Add content length on pendant request + Override Data

manoukianv:
This PR is to update de pendant request to add content-size in header to optimize http request and add override data.

This will be helpfull in IOT (arduino / stm32 / etc) project to reduce to get the data : when there is no size content, the iot frameworks wait 1s before get the data.

I update Pendant tools to build a bCNC panel (hardware).

### DIFF
--- a/bCNC/Pendant.py
+++ b/bCNC/Pendant.py
@@ -133,11 +133,11 @@ class Pendant(HTTPServer.BaseHTTPRequestHandler):
 						Image.open(tmp.name).save(out.name, 'GIF')
 						out.flush()
 						out.seek(0)
-						self.do_HEAD(200, content="image/gif")
+						self.do_HEAD(200, content="image/gif", cl=os.path.getsize(tmp.name))
 						self.wfile.write(out.read())
 				except:
-					self.do_HEAD(200, content="image/gif")
 					filename = os.path.join(iconpath, "warn.gif")
+					self.do_HEAD(200, content="image/gif", cl=os.path.getsize(filename))
 					try:
 						f = open(filename,"rb")
 						self.wfile.write(f.read())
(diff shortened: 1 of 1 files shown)


bCNC/Pendant.py
```diff
@@ -32,7 +32,7 @@
 	Image = None
 
 try:
-	import BaseHTTPServer as HTTPServer
+		import BaseHTTPServer as HTTPServer
```
Harvie: This indentation seems to be redundant to me...

bCNC/Pendant.py
```diff
@@ -89,26 +91,28 @@ def do_GET(self):
...
 
 		elif page == "/state":
-			self.do_HEAD(200, content="text/text")
 			tmp = {}
-			for name in ["controller", "state", "pins", "color", "msg", "wx", "wy", "wz", "G"]:
+			for name in ["controller", "state", "pins", "color", "msg", "wx", "wy", "wz", "G", "OvFeed", "OvRapid", "OvSpindle"]:
 				tmp[name] = CNC.vars[name]
+			contentToSend = json.dumps(tmp)
+			self.do_HEAD(200, content="text/text", cl=len(contentToSend))
 			self.wfile.write(json.dumps(tmp))
```
Harvie: probably no need to call json.dumps() again, perhaps we can reuse contentToSend?

Harvie:
Thank you! Seems sane, i will merge this. Can you please fix the cosmetic problems i've described?

bCNC/Pendant.py
manoukianv: You right, i will fix it

bCNC/Pendant.py
manoukianv: Yes, i fix it, it's not optimized.

Harvie:
Thanks. Please note that if you are developing some improved web pendant, there are already people working on this, so be sure to join efforts with them: #1121

PR #1252, opened 2019-07-14
Improvenets in plugins

apshu:
Hi, I am working on improving the bCNC. 
1) Improving the plugins. Addig documentation, improving functionality
2) Improving the Smoothie support.
Please accept the pull request.

Kind regards,
   Attila


bCNC/CNC.py
```diff
@@ -3362,6 +3362,17 @@ def drill(self, items, depth=None, peck=None, dwell=None, distance=None, number=
 			opname = "drill"
 		else:
 			opname = "drill:%g"%(depth)
+		#Incorrect peck values can block drilling cycle calculation
```
Harvie: Can you please in short explain this to me?

bCNC/ToolsPage.py
```diff
@@ -820,6 +820,21 @@ def __init__(self, master):
...
+
+* depth : Depth of the drill. If not provided, stock material thickness is used. (usually negative value)
+
+* peck: Peck step depth. If provided, drill with peck depth step, raising the drill to z travel value. If not provided, one pass drill is generated.
+
+* dwell: Dwell time at the bottom. If pecking is defined, dwell also at lifted height.
+
+* distance: Distance between drills if drilling alog path. (Number of drills will superceed this parameter))
+
+* number: Number of drills if drilling along path. If nonzero, Parameter 'distance' has no effect."""
```
Harvie: In the long run i would really like to merge drill from ToolsPage.py and Driller plugin. So we don't have two tools for drilling... (or three when i also count the drillmark plugin) I would be happy if you could help with this :-)

bCNC/plugins/flatpath.py
```diff
@@ -27,39 +29,70 @@ def __init__(self, master):
 		Plugin.__init__(self, master,"FlatPath")
 		self.icon = "flatpath"			#<<< This is the name of file used as icon for the ribbon button. It will be search in the "icons" subfolder
 		self.group = "CAM"	#<<< This is the name of group that plugin belongs
-		self.oneshot = True
```
Harvie: Does this really need parameters? I think this should be very simple tool which will create 2D projection of g-code in XY plane, no matter what. No need to have settings, we need to fix the code to work under all circumstances.

bCNC/plugins/flatpath.py
```diff
@@ -27,39 +29,70 @@ def __init__(self, master):
...
 
-
-		#active = app.activeBlock()
-		#if active == 0: active+=1
-		active=-1 #add to end
-		app.gcode.insBlocks(active, blocks, "Shape flattened") #<<< insert blocks over active block in the editor
-		app.refresh()                                                                                           #<<< refresh editor
-		app.setStatus(_("Generated: Flat"))                           #<<< feed back result
-		#app.gcode.blocks.append(block)
+		return blocks
```
Harvie: Please put newline at the end of the file.

bCNC/CNC.py
```diff
@@ -190,7 +190,7 @@ def load(self, filename=None):
 		def read(f):
 			while True:
 				line = f.readline()
-				if len(line)==0: raise
+				if len(line)==0: raise Exception()
```
Harvie: What is this for?

bCNC/CNC.py
apshu: Hi Harvie,
   If peck depth is negative input from the user, the pecking is working upwards from the material and not in the correct direction. The peck drilling while loop will never exit. (ln.3442)

bCNC/ToolsPage.py
apshu: How could I help? Do you think merging the three plugins into one?

Harvie: Yes. I think we can have single driller plugin to do everything. And it can also be used to generate drill marks instead of real drilling...

bCNC/CNC.py
Harvie: So why not just take abs() of the value, that would be much shorter code (= easier to maintain). Do you thing that would work?
Currently we have some issues with bCNC maintainability, so i would love to simplify things as much as possible.

bCNC/CNC.py
apshu: The Original implementation was generating a TypeError
  File "C:\projects\bCNC\bCNC\CNC.py", line 193, in read
    if len(line)==0: raise
TypeError: exceptions must be old-style classes or derived from BaseException, not NoneType

Now it generates a normal Exception. An assert would be better. I will change to assert, which reflects the usecase more.
  File "C:\projects\bCNC\bCNC\CNC.py", line 193, in read
    if len(line)==0: raise Exception()
Exception

Harvie: oh, i didn't saw there was `raise` already... makes sense. thanks.

bCNC/plugins/flatpath.py
apshu: The original "FlatPath" did not work for me. So I left it untouched with a wrapper option...

I see the "remove Z" and "Change Z" options as strategically different. 

The "remove Z" strategy is very useful for machines without a Z coordinate (eggbot or laser cutter). The side effect is the Z height of the 2D shape will be locked to previous shape's Z exit height. 

The "change Z" is flattening the path by setting all Z moves to same height. (stock material surface) This keeps the Z moves and the path will remain on the set level even if previous path's Z exit height is changed.

For the missing "slice path with a plane" strategy, I have not enough geometry knowledge. (Slicing a 3D tilted G02 is too much for me)

An additional option would be good to keep rapid movements at zsafe height. This is semi flattening...

If you decide to have one fixed flattening strategy, I would choose the "remove Z". 
Please let me know if I should commit a "oneshot" version.

Harvie: FlatPath is kinda development debugging feature, which simulates what happens if you re-cut g-code. For example you have circle and you cut it like helix with tabs. Then you decide to re-cut that with different step or flat passes instead, so bCNC needs to convert the helix back to flatpath and then apply the cut again.

This normally happens automaticaly, when you apply CUT on something that was CUT previously.

So there are two reasons why we have flatpath:

1.) It does exactly what CUT does with path before cutting, so we can check if this works correctly (actually there ARE issues with this, so i use flatpath to debug)

2.) To get original geometry from objects that have been previously cut (note if you have circular spiral with multiple passes, this SHOULD give you only single circle, not multiple circles in Z0)

Hope this makes motivation behind flatpath bit clear

bCNC/CNC.py
apshu: I was using the more complex style for those who have the stock material settings "upside down".
I can modify to abs() if you prefer that version.

bCNC/plugins/flatpath.py
Harvie: Also if you take a look on how flatpath works. This is the only important line:

`eblock = app.gcode.fromPath(app.gcode.toPath(bid)[0],eblock)`

Which means all the important stuff happens in `app.gcode.toPath()`, if there is some problem in flatpath, it needs to be fixed there, so it works in any corner case and internal representation of path is always properly constructed.

Harvie: What is your desired usecase for flatpath?

bCNC/CNC.py
Harvie: We don't really encourage people to use "Z down possitive" coordinate system. In my opinion this is wrong hardware setup and there are lots of things that will be broken in bCNC if you use this. It's like chinese machines comming with Y and Z swapped... Just fix the wiring :-)

bCNC/plugins/flatpath.py
apshu: Thanks for explaining the idea behind FlatPath. I am using path flattening for a different purpose. I am using free tools (QCAD, Inkscape, solvespace) for sketching the geometry and bCNC for CAM and control. (Hobby machinist...)

Flattening the imported path gives perfect WYSIWYG starting point for bCNC CAM tools.( Cut, offset, drill, etc. )

Feeding Z!=0 coordinates to my laser cutter, makes it angry :) I often found myself search-replace "remove Z coordinate". I was thinking automating this process is useful for others as well using bCNC.

Do you think it makes sense to keep all three options?

bCNC/plugins/flatpath.py
apshu: Implemented.

bCNC/plugins/flatpath.py
Harvie: I don't know, but importing both DXF and SVG should give you flat path by default... No need to flatten it anymore. This is my common workflow.

apshu: I was testing the "Unwind helical" feature with a simple test below, finding it semi-functional.

1) new file
2) CAM->Generator->Text
3) Generate text "Hello"
4) Select path "Hello"
5) CAM->FlatPath
Result is "H" while the other letters are missing from the flattened path.

If you split the generated text block, then flattening is working as expected.

apshu: The DXF import lacks rotation of objects. Even if it is n*90deg rotation. If you import a 15 deg rotated object, the drawing will look obviously incorrect. (QCAD exports these files) 
`dxf.py@521
	def rotate(self, r):
		"""Rotate entity by angle r"""
		error("Rotation of blocks is not supported for the moment""")`

I could not find any way to import the latest Inkscape SVG with correct scaling and without "extra markup objects" to bCNC and/or FreeCAD LibreCAD, etc. 

I found GCODE as the only really software independent format to transfer sketches to CAM.

apshu: By the way I get your point. When all bugs are fixed most of the features are unnecessary. 

Probably it would be good to add an independent post-processing plugin to "remove Z" coordinate for laser cutters and eggbot... And not mix it with the flattening feature. Calling it reduce parameter. Making universal for any optional parameter type. 

Removing all "Fxxx.xxx" from a block, or removing "E" or "Z" all makes sense.

Harvie: > The DXF import lacks rotation of objects.

Never needed that, but i guess you can add the support if you wish.

> I could not find any way to import the latest Inkscape SVG with correct scaling and without "extra markup objects" to bCNC

Please see https://github.com/vlachoudis/bCNC/wiki/SVG-support

> Probably it would be good to add an independent post-processing plugin to "remove Z" coordinate for laser cutters and eggbot

Have you tried bCNC in laser mode? I never used that, but if i remember corectly it was something like that... But i don't get why laser cutter complains about Z if you use GRBL it should be OK, just don't connect Z wires to any motor :-)

Anyway, i think it might be nice to have some search and replace (in selected blocks) functionality in bCNC. Especialy implemented as commandline command (which can be then also called by editor GUI). Idealy supporting both plain replacing and regexp replacing.


> I was testing the "Unwind helical" feature with a simple test below, finding it semi-functional.
> If you split the generated text block, then flattening is working as expected.

Yes, this is known bug. We should fix that in bpath.py, not in the flatpath plugin. I just wasn't priority for me... Also i didn't found elegant solution yet.

Harvie: Also... If you feel like you need to flatten 2D drawing that you've just imported, please make sure you have surface Z set to 0:

image

Harvie:
Can you please try flat_path from git master now? I've done some fixes to the underlying library. it's not perfect, but does allow flatpath to be applied on joined blocks without calling "split" first.

Issue #1238, opened 2019-06-06
WIth "Dark" theme,MAIN MENU BAR is Light GRAY with the TEXT in WHITE!!


lumanga:
Using a "Dark" theme, the MAIN MENU BAR is "Light GRAY" coloured  with the TEXT in WHITE.
In this case is impossible to read the text!

I have found a similar issue here:
https://github.com/vlachoudis/bCNC/issues/450

but witout a solution.. Can be now solved with version bCNC 0.9.14

ptvoinfo:
I've migrated to Ubuntu 18.04 and got the exactly the same problem :(.

Harvie:
Duplicate of #511

Issue #450, opened 2016-10-03
text color same as background color in GUI


Bodagous:
Hello, I do not think this is an issue with bCNC directly however when I choose a different theme in Ubuntu mate and than run bCNC the text is no longer visible in the drop-down boxes.  To be more specific the text is visible it is just the same color as the background of the boxes...(white on white = not visible) I'm using themes for gtk-2.0 and have tried changing the text color of the theme but still bCNC text remains not visible.  I'm not experienced with python at all and just a few years with Linux so my knowledge is still not enough to go digging through src files to attempt to remedy this on my own.(I tried anyway with no success). any help or point in right direction is appreciated. 

![screenshot_bcnc_textissue](https://cloud.githubusercontent.com/assets/20256625/19027109/ffab62fe-88e1-11e6-917b-24e7bf833f1c.png)

Effer:
I can't help you with this issue but I have to say that the black theme really looks cool!

PR #1283, opened 2019-09-24
System colors customization

ptvoinfo:
System colors customization (very useful for dark Ubuntu system themes).

### DIFF
--- a/bCNC/ControlPage.py
+++ b/bCNC/ControlPage.py
@@ -189,7 +189,7 @@ class DROFrame(CNCRibbon.PageFrame):
 		# work
 		col += 1
 		self.xwork = Entry(self, font=DROFrame.dro_wpos,
-					background="White",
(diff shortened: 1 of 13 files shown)


Harvie:
Can you please add some screenshots to explain what this does?
I've tried running it on XFCE with dark theme, but the bCNC UI is still in light colors.

ptvoinfo:
This problem is mentioned here (with the screenshot): #450, #1238 

The main problem is text in edit fields (white text on white background):
My screenshots (before/after with customized colors):

img1
img2

Harvie:
Thanks. I've merged it, but we still need to test it on other platforms. Eg. Windows.

ptvoinfo:
Sure. Therefore, The default INI file contains empty values. In this case, my update does not change anything. It'll take effect only if a user defines a color.

Issue #1063, opened 2018-11-23
I can not install the latest version (No ARM package for OpenCV in PyPI)


1f7:
I try install to Ubuntu 16.04 according to the last instruction with no luck
```bash
orangepi@OrangePI:~$ pip2 install --upgrade bCNC  
Collecting bCNC  
  Using cached https://files.pythonhosted.org/packages/fe/b2/96469773dfba552ea45dcbf2cacc255d5d1d651f6cfe8b0a75beb082d2cc/bCNC-0.9.14.39.tar.gz  
Collecting pyserial<=3.0.1 (from bCNC)  
  Using cached https://files.pythonhosted.org/packages/42/68/41bab8a5e3e50705f25358f043181242fa197f313aa0476908d9852c8dc4/pyserial-3.0.1.tar.gz  
Collecting numpy>=1.15.4 (from bCNC)  
  Using cached https://files.pythonhosted.org/packages/2d/80/1809de155bad674b494248bcfca0e49eb4c5d8bee58f26fe7a0dd45029e2/numpy-1.15.4.zip  
Collecting opencv-python>=3.4.2.17 (from bCNC)  
  Could not find a version that satisfies the requirement opencv-python>=3.4.2.17 (from bCNC) (from versions: )  
No matching distribution found for opencv-python>=3.4.2.17 (from bCNC)  
You are using pip version 8.1.1, however version 18.1 is available.  
You should consider upgrading via the 'pip install --upgrade pip' command.
```
Previous versions were installed without problems.
opencv-python need compile by hand?

**P.S.**
git clone works fine, both @vlachoudis  and @Harvie

1f7:
> Does `sudo apt-get install python-opencv` help?
```bash
 sudo apt-get install python-opencv
Reading package lists... Done
Building dependency tree
Reading state information... Done
python-opencv is already the newest version (2.4.9.1+dfsg-1.5ubuntu1.1).
The following packages were automatically installed and are no longer required:
  libllvm3.8 libmircommon5
Use 'sudo apt autoremove' to remove them.
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```


> Does `sudo pip2 install opencv-python` work?
```bash
 sudo pip2 install opencv-python
The directory '/home/orangepi/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
[...]

PR #1285, opened 2019-09-24
Partially improved the Russian translation

ptvoinfo:
Partially improved the Russian translation
Added arm64 exclusion for opencv in setup.py


Harvie:
Thanks! I wonder if this will fix #1063 (i've uploaded it to pypi)

Issue #1209, opened 2019-04-06
Prints Small


KenwoodFox:
All of my paths show up at one tenth the size, my board interperts mm, my gcode is in mm, my cam software is in mm, why does bCNC seem to move my pinter with the decimal point in the wrong spot?

Harvie:
Go to bCNC settings and check if you have mm or inches...

KenwoodFox:
im reading 1/10 not 1/25 but ill check those settings, and yes bCNC has that dropdown selecting mm, are those the settings you're referring to?

Harvie:
What input file do you use? Are you loading g-code to bCNC?
If loading DXF, SVG, STL or PLY, you should make sure that it's in mm, not cm...

paulvdhoeven:
I'm feeling a bit silly for mentioning it, but are you sure you've callibrated the number of steps /mm properly in Grbl? Maybe you accidentally entered calibration settings for cm, and are sending mm's to it.

Teddyz:
I see a similar problem when saving as svg from Inkscape. Everything becomes small in bCNC. When I instead export as dxf from Inkscape it gets correct size.

See example:
![bild](https://user-images.githubusercontent.com/14166139/65354019-cf696100-dbee-11e9-85d9-c68e2bda734a.png)

[elt.zip](https://github.com/vlachoudis/bCNC/files/3637463/elt.zip)

But I am a rookie here. Any recommendation where I can post questions?

Harvie:
@Teddyz what DPI you use for inkscape export? Do you have your bCNC and GRBL correctly set in metric?

Teddyz:
@Harvie I do not know how to see the DPI setting you ask for. Can it be seen in the saved files? They are attached in my message above. 

I think I have bCNC and GRBL set OK regarding metric settings. I have installed switches and Homing works. Movement and display of xyz-values are OK and I can use the router. 

I posted because I have similar problem as OP, but my size-difference is not exactly 1:10. I use bCNC 0.9.14-dev.

Harvie:
Have you read this? https://github.com/vlachoudis/bCNC/wiki/SVG-support
You might need to apply transforms before exporting as SVG for use with bCNC
There is issue open for this: https://github.com/regebro/svg.path/issues/36

KenwoodFox:
[...]

PR #1312, opened 2019-12-02
svg parsing via svg.elements

tatarize:
Move the SVG parsing to svg.elements project ( https://github.com/meerk40t/svg.elements ). This should provide highly robust SVG parsing, much more inline with the SVG spec. This should correct the or make correctable the transformation errors with loading SVG.


tatarize:
See #1209 for more why this change is needed.

Harvie:
Hello, thanks for your effort. I will surely test and review the code this week. In the meantime i have few question to get better understanding of the big picture:

Would you consider this code reasonably stable? (i hope we can release new stable bCNC soon)
Are there any downsides of using this vs svg.path?
Is svg.elements based on svg.path or it's complete rewrite?

Harvie:
So i've done first test and it really seems that it can import stuff that can't be imported by my previous code.

But it seems to me that all elements from SVG are imported TWICE. why is that?

eg. this random curve i did in inkscape produced two SVG blocks, one of them highly subdivided, one of them quite rough:

image

What causes this issue?

Harvie:
Also this random SVG i've pulled from google seems to take much longer to import and produce less effective g-code (more lines) than it does with original import code.

[delete.svg.txt](https://github.com/vlachoudis/bCNC/files/3913832/delete.svg.txt) (sorry for the image content, i selected it because the gradient felt like it might cause troubles with parsing, not for the hammer figure :-)

Harvie:
Also the code does not run at python2:

```
$ python2 -m bCNC
new-config Utils <ConfigParser.ConfigParser instance at 0x7fb8df880910>
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/harvie/Temp/tatarize-bCNC/bCNC/__main__.py", line 58, in <module>
    from CNC import WAIT, CNC, GCode
  File "/home/harvie/Temp/tatarize-bCNC/bCNC/CNC.py", line 25, in <module>
    from svgcode	import SVGcode
  File "/home/harvie/Temp/tatarize-bCNC/bCNC/lib/svgcode.py", line 16, in <module>
    import svg.elements
ImportError: No module named elements
```


I guess this is caused by the module structure. See differences between directory structure:

bCNC/lib/svg/elements/__init__.py
and
bCNC/lib/svg/path/__init__.py

tatarize:
> Would you consider this code reasonably stable? (i hope we can release new stable bCNC soon)
Are there any downsides of using this vs svg.path?

It is pretty reasonably stable. I've been working on crushing minor edge condition bugs that otherwise would never have parsed. It does a lot of the svg 1.1 test suite perfectly now.  I don't see any downsides to this code over svg.path. The one caveat I would note is that I've been working on it a bunch for a bit now. ( https://github.com/meerk40t/svg.elements ). It's had a number of updates in the last few weeks. Most of these to cover edge conditions and specific sections of the standard. Going forward, I'm still going to do some of these corrections. Replace the Distance class with the lazy interpolated Length class, perhaps get the defs and css styling working. While it works now for what it does, and doesn't have any known bugs, I will still be improving it a bit for a while. So it might be worth an update to the code in a month or so, even though the differences would be very minor comparatively. And you might never run across some edge condition that failed.

> Is svg.elements based on svg.path or it's complete rewrite?

A bit of both. It's based on svg.path but is such an extensive rewrite that it's hard to find the code that remained the same. Mostly it's backwards compatibility with svg.path and using the same methodologies as regebro used. And keeping the objects compatible like point which works like a complex even though it's not and can do more work. Some changes went the other way as I gave him my solution to the subpath problem. I added and tested against his entire test suite for the code and the only failing tests became issues since the problem was in his didn't fit the spec.


>But it seems to me that all elements from SVG are imported TWICE. why is that?

Silly bug. The second if statement needs to be an elif statement since the Path is also a Shape. 

>eg. this random curve i did in inkscape produced two SVG blocks, one of them highly subdivided, one of them quite rough:

It's likely due in part to the size or rounding as such. The shapes produced aren't subdivided natively. They are merely handed off to the code that was already there. And I'm not sure how often those elements divide such things.

> I guess this is caused by the module structure. See differences between directory structure:

The structure for the library is a bit strange. I had some trouble getting to it to work initially. I think it might be better to remove that svg subpath thing and include it without those oddities to try to get that dot there. It was done is svg.path so I copied that. It is weirdly finicky.

tatarize:
Moved the svg_elements.py file to the lib directory. Updated the code to avoid the double bug. I think the much larger breakup of the shapes is due to some issue of scaling.

Also, keep in mind the delete.svg you sent actually does outline the shape multiple times it does this to add the gradient fills to them. I'm not sure what the decomposition into points should be. Maybe I messed up with the scaling factor and one of the divisors was to stop it from overly subdividing.

tatarize:
I too am maintaining python2 support until End of Life 1/1/2020. So I'm hoping to get it into a solid position where it's capturing most of the edge elements in the spec before then. Then I'll just make a python2 branch and discontinue support. Until then it'll remain python 2 and 3 compatible. Getting python2 and 3 importing working right has always been a massive chore, and I'm sorta envious of fstrings.

tatarize:
Okay, that should fix the issues you raised. Turns out, oddly enough:

`shape = type(segment).__name__`

Isn't backwards compatible since my classes don't come from object and so they all claimed to be 'instance'. Somehow since I use instanceof() so often that methodology slipped my grasp. I'll look into it since I try to be strongly backwards compatible. I pushed your scale factor into the code correctly as a modification of the typical PPI. I found the statistics button and figured out the correct scaling. I also set the default size to be 100 for unset length and width, which should be a bit more forgiving.

Those two bug fixes combined makes it properly nice. Give it another shot. And if you need to rebase the stack (some people hate multiple commits with odd names) you can just overwrite svgcode.py and add in svg_elements and everything should work fine.

Harvie:
OK. This looks nice. I can even import SVGs that were broken in previous version!

I have probably one last problem with this:
`svgcode.get_gcode()` ignores the `subdivratio` setting passed from `bCNC/CNC.py`.

In my original code i am able to tune subdivision using that parameter, but it does not work in your code. Default value is 10, which means that spline of length 1 will be subdivided to 10 lines. If i increase the number i get more detail, if i decrease it i get less detail (and shorter g-code).

I can't do that with your code. This is kinda important to me, because bCNC has currently troubles when handling too long g-code files, so we need way to tune this. I even want to expose this setting to users...

tatarize:
Yeah, I lost the propagation of that variable. Both digits and subdivratio were there for the Shape but not for Path. That's actually why you got a rough and a less rough version of the same path. It's a quick update.

tatarize:
This should be able to load basically all SVGs there's a few things like `<defs>` and `<use>` and the ability to do CSS styling that isn't fully covered yet. When I get it covered I'll send up another pull. For now it'll be very significantly more robust. You could even assign meaning to the stroke or fill colors if you wanted. Like red=some-speed, blue=other-speed. Not that it's that interesting but it's how lasering works as such. Basically all the bits are properly processed, so if you wanted them for something...

Harvie:
Thank you. Seems to be working nice. I will merge this in few moments.

BTW I have to lower subdivratio in bCNC from 10 to 0.5, because now that scaling works, it seems to produce too much details (20 times more g-code lines). But i guess that's to be expected... I will try to make this configurable anyway...

Harvie:
BTW What about svg.elements licensing? Is it still MIT code? If so can you please add your personal details to LICENSE.MIT file in a same manner that svg.path is mentioned there?

tatarize:
Yeah, all the code is still MIT licensed. I'm not worried about it, the comment at the top of the code makes it clear that the code is MIT licensed and that it derives from some other code which is equally MIT. Some bits of Zingl's brilliant work and svgpathtools and svg.path projects. If you need my name you can toss it there as `Copyright (c) 2019 David Olsen.` Zingl was the only person who wasn't super-overt with his licensing and I had to email him for clarification. The SVG proper stuff doesn't need any of his code and I've considered turning it into a different module. It basically does bresenham's line draw algorithms for all the curves so it does pixel perfect shape drawing.

 Apparently the code base has the copyright listed as meerk40t ( https://github.com/meerk40t/svg.elements/blob/master/LICENSE , did it automatically and never bothered to look). I'll add the license in the svg_elements.py file directly so if I get around to implementing the `<defs>` correctly and you toss in a pull request, it'll fix it there too. 

I don't think anything in else you have depends on svg.path anymore, and it could be safely deleted.

Issue #1314, opened 2019-12-06
upgrade then it does'nt run anymore


dhuvelle:
Hello everyone, 

I'm have troubles since i upgraded with "python -m pip install --upgrade bCNC". I'm no more able to launch bCNC and it says me the following lines

`C:\Users\dhuve>python -m pip install --upgrade https://github.com/vlachoudis/bCNC/archive/master.zip
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
Collecting https://github.com/vlachoudis/bCNC/archive/master.zip
  Downloading https://github.com/vlachoudis/bCNC/archive/master.zip
     | 7.2MB 4.9MB/s
Requirement already satisfied, skipping upgrade: numpy>=1.12 in c:\python27\lib\site-packages (from bCNC==0.9.14.301) (1.16.5)
Requirement already satisfied, skipping upgrade: Pillow>=4.0 in c:\python27\lib\site-packages (from bCNC==0.9.14.301) (6.2.1)
Requirement already satisfied, skipping upgrade: pyserial<=3.0.1 in c:\python27\lib\site-packages (from bCNC==0.9.14.301) (3.0.1)
Requirement already satisfied, skipping upgrade: opencv-python>=2.4 in c:\python27\lib\site-packages (from bCNC==0.9.14.301) (4.1.2.30)
Installing collected packages: bCNC
  Found existing installation: bCNC 0.9.14.301
    Uninstalling bCNC-0.9.14.301:
      Successfully uninstalled bCNC-0.9.14.301
    Running setup.py install for bCNC ... done
Successfully installed bCNC-0.9.14.301

C:\Users\dhuve>python -m bCNC
new-config Utils <ConfigParser.ConfigParser instance at 0x0000000002DED408>
================================================================================
WARNING: bCNC has been recently ported to support both python v2.x and v3.x
Most things seem to work reasonably well in both python versions.
Please report any issues to: https://github.com/vlachoudis/bCNC/issues
================================================================================
self.canvas.winfo_id(): 17435774
Using old style comports()!
Using old style comports()!
Using old style comports()!
[...]

PR #1328, opened 2019-12-28
Some BugFix following python3 port

DodoLaSaumure:
ops is filter type in python3 , convert to list


Harvie:
Will this work in both python2 and python3? We still have python2 users...

DodoLaSaumure:
@Harvie yes it works both on python2 and python3, just tested.

DodoLaSaumure:
commit 25fb511 should correct #1314 https://github.com/vlachoudis/bCNC/issues/1314

DodoLaSaumure:
Commit 2e41fc5 & a672e1c : 
python3 pickler does only accepts ByteIO, not StringIO, and I did not find anyway to convert Block dump tuples to bytearray. I found it more easy and more pythonic to wrap data in json objects.

DodoLaSaumure:
all commits of this branch tested on both python2 & python3

Harvie:
`import json` is internal part of python or does it need to be installed as bCNC dependency?

bCNC/lib/tkExtra.py
```diff
@@ -3075,13 +3076,18 @@ def afterHide(self):
 	# Public methods
 	# ----------------------------------------------------------------------
 	def get(self, first=None, last=None):
+		PY2 = sys.version_info[0] == 2
```
Harvie: i would prefer python version checks not being used in code. is there another way to solve this? like better use of isinstance() ?

DodoLaSaumure:
> `import json` is internal part of python or does it need to be installed as bCNC dependency?
json is internal part of python3
https://docs.python.org/fr/3/library/json.html
and python2
https://docs.python.org/fr/2.7/library/json.html

bCNC/lib/tkExtra.py
DodoLaSaumure: Thanks @Harvie checkout commit 723bb86 please

Harvie:
OK thanks, seems resonable... Do you have something more going on, or should i merge?

DodoLaSaumure:
Thank you @Harvie Please merge !! And all the best for this end of year !!

Harvie:
Copypasting now works for me on both py2 and py3. Thank you very much!

Harvie:
@DodoLaSaumure can you please take a look at this? #1336
There seems to be some problem with json clipboard...

PR #1329, opened 2019-12-28
Update tkExtra.py

DodoLaSaumure:
corrects a bug in displaying locals with python2, works with python3 also


Harvie:
`.decode()` might be more readable way doing this...
BTW You can just push more commits to #1328 branch...

PR #1345, opened 2020-01-27
Add new line to sender so Smoothieware can be used over TCP

wsidebottom:
Add new line to sender so Smoothieware can be used over TCP

### DIFF
--- a/bCNC/controllers/_GenericController.py
+++ b/bCNC/controllers/_GenericController.py
@@ -84,7 +84,7 @@ class _GenericController:
 		self.master.sendGCode("$H")
 
 	def viewStatusReport(self):
-		self.master.serial_write(b"?")
+		self.master.serial_write(b"?\n")
 		self.master.sio_status = True
 
 	def viewParameters(self):


Harvie:
Seems like buffer issue to me. Have you tried calling `self.master.serial.flush()` instead of sending newline?

wsidebottom:
Just tried that. it only likes it if a newline is sent, otherwise it says 'Not Connected'.
I have been using bCNC for a few years with this mod with different versions of Smoothieware over TCP.
This issue is only present when TCP is used not serial over usb.

Harvie:
Well. I am not strictly against this. However i have some things that IMO should be figured out.

1.) we need to test if this does not cause problems with other setups like GRBL over serial.
2.) we might be able to autodetect that we run over network (eg. from pyserial url) and send the newline only in such case. Do you use the pyserial url handler to make the network connection or something else?

wsidebottom:
I use socket://<IP ADDRESS>:23
I dont have grbl installed on anything to test it but i do know it does not affect smoothie over usb with this mod.

wsidebottom:
Perhaps you could test it with your setup and see if there are any issues.

wolfmanjm:
This will cause issues on Smoothie over USB. It will cause an extra unwanted ok to be returned, which will get the ok count out of sync.

Harvie:
> This will cause issues on Smoothie over USB. It will cause an extra unwanted ok to be returned, which will get the ok count out of sync.

This is a problem. I will close this PR for now, but i am looking forward to new PR which will not have this problem. Thanks for improving bCNC guys.

wolfmanjm:
FYI the latest version of smoothie does not need the \n for ? anymore over network, so it works the same as for USB connections. (For now this is not in the prebuilt smoothie firmware, so build from source).

Issue #1256, opened 2019-07-26
Dialog Box in Button script?


BernardG:
As many others, I would like to write a script on a button to probe on X, Y and Z. The thing is, as you know, if the script is pure GCode, I will need to use one button per tool diameter. (not even talking about choosing which corner to probe)

So, I wonder if there is a way to use some python code to ask the user his tool diameter, and use the variable value in GCode?

I know about the possibility to write plugins, but that does not look (to me) as a plugin job, should be juste used on one button, if at all possible.

Thanks for a great piece of software!

Harvie:
Never tried this, but from my understanding, this should be possible according to https://github.com/vlachoudis/bCNC/wiki/Scripting

First, you have to asign variable like this:  `%_yourvariable=123` then you can use it in g-code like eg.: `G0 X10 Y#yourvariable`. You can replace the asignment with any arbitrary python line, so you should be able to make python popup that asks for a value to asign to _yourvariable.

Harvie:
So basicaly i've found workaround for you and mentioned in wiki. Code goes like this:

```
%_myval=raw_input("enter value: "); print("entered value: "+_myval);
G0 X10 Y[_myval]
```

![image](https://user-images.githubusercontent.com/276504/61949059-e6525500-afa9-11e9-90c1-61414ab27b08.png)

In future we can upgrade this to provide popup, but this should be good enough to get the work done for now...

BernardG:
Yes, that will be good enough for now. Thank you so much! Sure a popup would be "more classical" and look nicer, but this will get the job done, which is what really matters.

Harvie:
I've just tested and it also works in buttons:
![image](https://user-images.githubusercontent.com/276504/61967379-6f32b600-afd5-11e9-898d-691728698216.png)

What OS do you use? It's important to open bCNC from commandline.

BernardG:
I have been trying with Ubuntu Linux, not yet from the command line in Windows 10.
Where do you enter asked value? On bCNC command?

Harvie:
Enter the value to the terminal and hit enter.

BernardG:
OK, got it. It's just such an ugly workaround that I did not even think that was the way to go! :-)
Now, what would be the proper way yo call a python popup instead?

Harvie:
@BernardG don't worry, i think i might roll some new code for this today... i am very close!

Harvie:
I got it working, will upload in minute:

[...]

PR #1378, opened 2020-04-04
Scripting now always has access to app

### LINKED ISSUE #1375: Controlling GRBL from python scripting
I'm trying to control GRBL from bCNC python scripting. I'm trying to interleave actions controlled with the gpio on the rpi running bCNC with GRBL movements.

I am using the python scripting successfully however I am stuck on controlling GRBL.
I thought I should be able to execute GCODE from inside the scripting but I can't find a way to do this... 

From the commandline app.executeGcode works, but when the same code is executed from a gCode file, app is not available. The wait command also doesn't work.

Am I going about this the wrong way? 

Further info:
I have a tool changer hooked up to the rpi which is working but I need to be able to do one step with the changer, then move, then another step. Ultimately I would like to be able to generate my gcode files to include the tool changing script to have fully automated runs. I may also integrate bit inspection...

Any help appreciated.
Thanks

flemish4:
Fixes #1375
This change:
Improved scriptability #1256
https://github.com/vlachoudis/bCNC/commit/4456ad2abe63121899e4833f0b3b04323d6e278c

Added the ability for scripting to access 'app', however it only added this for the commandline.
The pull request adds the functionality in two other places.

- sender.py:200           - commandline
- CNCCanvas.py:1817 - opening gcode and rendering
- sener.py:710             - running

### DIFF
--- a/bCNC/CNC.py
+++ b/bCNC/CNC.py
@@ -1091,8 +1091,9 @@ class CNC:
 			else:
 				try:
 					return compile(line[1:],"","exec")
-				except:
-					# FIXME show the error!!!!
+				except Exception as e:
+					print("Compile line error: \n")
+					print(e)
 					return None
 
 		# most probably an assignment like  #nnn = expr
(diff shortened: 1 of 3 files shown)


Harvie:
thanks

PR #1396, opened 2020-05-05
Update svg_elements.py

### LINKED ISSUE #1395: No G03 codes in .SVG import.
When opening wings.svg in bCNC 0.9.14-dev: (With Python 3.6.9 on Linux Mint Xfce) all the arcs in a path are read as G02 codes, while 10 of them should be G03.
The file was exported from FreeCAD 0.18.16146
When the file is opened in Xviewer or Inkscape it looks as expected.
The coordinates are OK, When I manually change the G02 codes to G03 in bCNC the path is good.

[wings.svg.zip](https://github.com/vlachoudis/bCNC/files/4541945/wings.svg.zip)

DodoLaSaumure:
fix  #1395
works both with py2 & py3 (tested)

### DIFF
--- a/bCNC/lib/svg_elements.py
+++ b/bCNC/lib/svg_elements.py
@@ -3604,7 +3604,7 @@ class Arc(PathSegment):
 
         self.prx.matrix_transform(rotate_matrix)
         self.pry.matrix_transform(rotate_matrix)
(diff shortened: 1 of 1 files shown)


Harvie:
thanks.

PR #1408, opened 2020-05-27
Reduce halftone plugin output steps with conical mill 

m-g-m:
Patch to halftone plugin adds conditional to skip motion commands for spots below min size when the conical mill option is selected. Behaviour mirrors the main "circle" code conditional and results in significant reduction in output for sparse images.


Harvie:
Thank you very much.

PR #1434, opened 2020-07-14
CAM plugin for engraving text

apshu:
Hi Harvey,

    Please consider including this generator plugin. The plugin generates toolpath for engraving. The font is selectable and very simple in design. The module uses Hershey-Fonts to render text. The simplex typefaces are always single pass 'stick fonts' suitable for laser and traditional engraving.

   The plugin needs Hershey-Fonts from PyPi

Kind regards,
     Attila


Harvie:
There already IS text engraving functionality in bCNC. I dont like idea of adding second one.
Why is your better and why cant you just improve the existing one?

Harvie:
Sorry, but i will not merge this unless you explain why do we need to have two separate text generator plugins.
Or preferably add your functionality to the existing plugin.

Thanks for your efforts.

PR #1384, opened 2020-04-15
Initial 6 axis Support

tsmarks:
Added basic 6 axis features: DRO, Control
Added a 4th axis roughing strategy

As requested by creator, the DRO and Control panes can be removed/activated in the bCNC.ini file line:  control.page = DRO abcDRO State Control abcControl

PR #1441, opened 2020-08-09
Fix POSPAT & STATUSPAT to match non-6-axes GRBL output

sttts:
My 4-axes GRBL outputs `[PRB:-356.0,-5.28,-118.3,0.0:1]` for the probe position. This didn't match the 6-axes pattern (and not even the 3-axes from before https://github.com/vlachoudis/bCNC/pull/1384).


Harvie:
Nice. Please test on both python2 and python3 (probably not going to matter as it is only regex). 
Can you test it on multiple GRBLs? Each with different number of axes? I think the 3 axis setup is still the most used, so we should be sure it will be working properly.

sttts:
I tested different numbers of axes via https://pythex.org/. I don't have a real non-4-axes GRBL. The right way to avoid regressions would be unit tests, run for both python 2 and 3 as Github actions. Doesn't look like this repos is heavy on tests though :-)

Harvie:
> Doesn't look like this repo is heavy on tests though :-)

PR welcome ;-)

sttts:
🎉  first successful tool change today (after 5 or 6 emergency stops 😱).

Harvie:
Thanks and congratulations! Are you doing the toolchange manualy? It should not cause any emergency stop as far as i remember...

sttts:
Yes, manual change. And I did the emergency stops myself, pressing the big red button before the tool would crash into the ground. Note that due to the wrong patterns, the `mz` and `prbz` variables had random or zero values.

Harvie:
Oh i see. This part of code seems to be messy. I couldn't even agree on how it should behave in the first place, beacuse there seem to be several conflicting usecases regarding various offset variables and their combinations. I haven't managed to wrap my head around it yet... Any help would be appreciated.

sttts:
Actually the tool length procedure and the calibration func are pretty straight forward. A little more inline comments would help though.

IMO it's worse that it is not documented in detail what it actually does. There is the wiki page describing it, but eventually I had to turn to the source code to get the complete picture. Half of the buttons on the probe pages have zero docs.

PR #1445, opened 2020-08-17
Quick changes to make GRBL1.py not crash on machines with 4 or 5 axes.

tmbdev:
This get rids of the index errors in GRBL1.py when a machine has 4 or 5 axes. It introduces no new functionality otherwise and should be a save patch to apply.


Harvie:
Thanks

PR #1452, opened 2020-09-01
Add POSPAT tests

sttts:


sttts:
@Harvie I think travis build with 2.7 is broken independently from this PR.

Harvie:
Hello,
can you please explain the changes to me?

Also it does not even start:

```
$ python -m bCNC
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/harvie/Temp/bCNC-sttts/bCNC/__main__.py", line 50, in <module>
    import Utils
  File "/home/harvie/Temp/bCNC-sttts/bCNC/Utils.py", line 41, in <module>
    if getattr( sys, 'frozen', False ):
NameError: name 'sys' is not defined

```

Harvie:
Also there is already some similar change done by you merged in the past... Are you aware of this?

https://github.com/vlachoudis/bCNC/commit/5abe8cb98603f84ac57fdd67307e18c2b611796a#diff-4857d30b496749a712eae2d5bc42b2b5f8473e34daed6fffb46515519ca30ff4

Harvie:
Also what do you think about changes proposed in #1525 ?

aib:
This makes the 8th and 9th capture groups of the regex non-capture, i.e.:

`[PRB:-356.0,-5.28,-118.3,4.0,5.0,6.0:3]`

The `:3` and `3` (overlapping) parts just before the `]` in the example above, provided in the excellent tests.

As there is no code currently using groups beyond the 4th one (apart from the tests here), this should not affect working code.

Harvie:
> As there is no code currently using groups beyond the 4th one (apart from the tests here), this should not affect working code.

For me it is important for code to match the following requirements:

  * Be robust enough not to fail when some new grbl fork adds more data to the list (just read what we understand and ignore the rest)
  * Allow for multiaxis setups, no matter if 3,4,5,6 or more axes. There is already basic 6-axis support in bCNC (have to be enabled in config to show the GUI).

aib:
This PR will not change any code. It's just tests for POSPAT.

The other one allows 3-axis controllers to be used.
The other one is a fix for 5abe8cb98603f84ac57fdd67307e18c2b611796a which did not work as intended.

Harvie:
> This PR will not change any code. It's just tests for POSPAT.

What do you mean? Have you clicked on the "changed files" ?

image

Harvie:
> The other one allows 3-axis controllers to be used.

I still don't fully understand the problem... 3-axis GRBL seems to work with current bCNC code...

> They can be combined, though it might have to be done manually.

Are there any benefits from doing so except for the tests?

aib:
> > This PR will not change any code. It's just tests for POSPAT.

> What do you mean? Have you clicked on the "changed files" ?

I mean it will not change how the program behaves. It's a test, code intended for developers, not users. (And if you mean POSPAT I've explained the change and why it shouldn't matter above)

> I still don't fully understand the problem... 3-axis GRBL seems to work with current bCNC code...

GRBLv0? Not mine. Can you (or someone else) post an example status or position line?

> > They can be combined, though it might have to be done manually.
> Are there any benefits from doing so except for the tests?

Seeing as how the whole point of this one is tests, no :)

Harvie:
> GRBLv0? Not mine.

Oh i didn't noticed your PR is targeted to GRBLv0. In my opinion it is rather weird to use GRBLv0 nowadays, when v1 is already becoming obsolete...

That just brings another question... Are you sure your modifications targeted to v0 will not break compatibility with v1 protocol?

Harvie:
Maybe we should strictly separate patterns for individual GRBL versions, so we don't have to worry about this.

PR #1453, opened 2020-09-01
Add requirements for Mac

sttts:
This allows development inside an IDE like IntelliJ Idea on Mac.


Harvie:
Thank you

Harvie:
On the other hand. I am not really sure if we should install this to all users. Developer can install this individualy, but there is no need to install this for users i guess??

sttts:
I bet you cannot run it on Mac without those.

PR #1459, opened 2020-09-16
SVG subdivision ration pop-up dialog

m-g-m:
Added some code to pop up a dialog for SVG subdivision ratio, and save the value to the config file as default for next time. Trade off for this is user must respond to this dialog each time before the SVG file is loaded.


m-g-m:
WIll clean this up and resubmit

Harvie:
I don't think popup is good UX. There should be setting in bCNC config for this.

m-g-m:
Ok.
So far in my experience different files need different settings so I found the pop-up is more convenient than going into the config settings repeatedly.
Perhaps a more intelligent algorithm is a better long term answer.
Regards,
Matt

On September 16, 2020 11:27:57 PM MST, Tomas Mudrunka <notifications@github.com> wrote:
>I don't think popup is good UX. There should be setting in bCNC config
>for this.
>
>-- 
>You are receiving this because you modified the open/close state.
>Reply to this email directly or view it on GitHub:
>https://github.com/vlachoudis/bCNC/pull/1459#issuecomment-693967227

-- 
Matt Martin

Harvie:
I set the subdivision length to match precision achievable with my CNC machine. No need to use more precision than my machine can physicaly handle.

m-g-m:
What i run into is the files getting unworkable large, especially if I want to run some computations such as a profile on them.
Maybe that is just my situation.


On September 17, 2020 8:27:43 AM MST, Tomas Mudrunka <notifications@github.com> wrote:
>I set the subdivision length to match precision achievable with my CNC
>machine. No need to use more precision than my machine can physicaly
>handle.
>
>-- 
>You are receiving this because you modified the open/close state.
>Reply to this email directly or view it on GitHub:
>https://github.com/vlachoudis/bCNC/pull/1459#issuecomment-694312239

-- 
Matt Martin

Harvie:
Have you tried running "arcfit" command on the shapes before further processing? It will replace small subdivisions by circular arcs wherever possible to do it within defined precision. I plan to do this automaticaly during SVG import but i guess it has not been 100% tested yet.

m-g-m:
No, have not tried it - sounds promising !
Thanks,Matt
-----Original Message-----From: Tomas Mudrunka <
notifications@github.com>Reply-To: vlachoudis/bCNC <
reply@reply.github.com>To:
vlachoudis/bCNC <bCNC@noreply.github.com>Cc: Matt Martin <
matt.martin@ieee.org>, State change <state_change@noreply.github.com>Su
bject: Re: [vlachoudis/bCNC] SVG subdivision ration pop-up dialog
(#1459)Date: Fri, 18 Sep 2020 09:20:26 +0000 (UTC)

Have you tried running "arcfit" command on the shapes before further
processing? It will replace small subdivisions by circular arcs
wherever possible to do it within defined precision. I plan to do this
automaticaly during SVG import but i guess it has not been 100% tested
yet.

PR #1460, opened 2020-09-17
SVG subdivision ratio dialog when opening svg file

m-g-m:
Adds a dialog to set the SVG subdivision ratio when opening an SVG file. Ratio is saved to the config file to be "remembered" for next time. Note this changes usage flow requiring the user to close this dialog before an SVG file is imported.


Harvie:
Hello,
generaly speaking i don't like this idea. bCNC already has established ways to cope with such situations. These are:

1.) Provide configuration option somewhere in bCNC configuration menu
2.) Improve the SVG import code in order to autodetect sensible value automaticaly (might not cover all edge cases, so possibility of manual override would be considered vital)

I guess i will close this PR for now, because i don't like to annoy users with popups unless absolutely necessary. I think we should to consider other options before doing this. The user experience of bCNC is already rather sub-optimal and i think we need to improve the UI rather than clutter it with more popups.

I am sorry and even though i currently postponed inclusion of this feature i still do appreciate your effort and i beleive we can come up with some more fit solution.

PR #1490, opened 2020-11-16
Update README.md

bosd:
Default to python3 installation

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,19 +22,19 @@ This is short overview of installation proces, for more see the ![bCNC installat
 This is how you install (or upgrade) bCNC along with all required packages.
 You can use any of these commands (you need only one):
 
-    pip2 install --upgrade bCNC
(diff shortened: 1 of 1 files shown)


Harvie:
Thank you. Can you please update wiki a well? you should have access to it.

bosd:
I actually updated the wiki before I edited this file. :)

However some pictures in the wiki might need an update. I am not a windows user, but will try to find some time to update those.

As a new user I had some trouble with tkinter. Preferably I would like to have a warning from pypi when this requirement is not installed. Following up with instructions how to install it on the distro.


For further migration to python3, some code changes may be desired. Generally it is better to have python3 as default and Python 2 as a fallback.
(Now it is the other way around) This will make the phase out of Python 2 in the future easier.

Harvie:
> Generally it is better to have python3 as default and Python 2 as a fallback.

Agreed

PR #1500, opened 2020-12-23
Upgrade svgelements support & correct svgcode.py

tatarize:
Updates and corrects svg support.

There's some confusion in the use of the PPI value. In Inkscape this is 96.0 but in Illustrator this would be 72.0. This information isn't in the file itself, it needs to be set. Illustrator files would scale off by 25% without exposing the ability to correct that to the user. You can't just scale any document by that value because the document itself might give it's own units. The 0.5 and 1 values for `subdivratio` had no coherent units and these values were likely determined by trial and error. It's now samples per unit distance. Since the values are scaled to 1 unit = 1 inch or 1 unit = 1 depending on the `SVGScale()` value.

* Updates `svgelements` from about ~0.7.x to 1.4.1
     * Corrects several edge conditions in SVG processing.
     * Use and Defs flags
     * Viewport corrections.
     * Corrected CSS styling,
     * Better support length/percent values in basic shapes.
     * `numpy` accelerated `npoints()` function.
* Updates `svgcode.py`
     * Correct treatment of the conversion of SVG -> Gcode Arc.
     * Use of faster parsing functions for linear interpolated npoints() function
     * Corrected units on subdivratio. Now has understandable units. 200 samples per inch.
     * Correct segment.sweep value in Arc code. (This was a change between `svg.path` and `svgelements`) and was an oversight in the previous code.
     * Corrected PPI unit values vs. scale. Some svg files give their units in the `svg` root elements and thus it's incorrect to scale in the manner the code was scaling.
* `svg.path` deleted. This is an ancestor of `svgelements` code, svgelements is a superset of that code, and the direct use of that code was removed when the `svgelements` was initially added. 

UPDATE: No longer an issue. It's all fine.
Caveats: Might need to have a cursory check done with python 2.7. Support was dropped about a year ago at Python 2 End-Of-Life, and while nothing massive has changed, there's some chance basic parsing might require a tweak. Dropping support means I just haven't checked.


tatarize:
Updated svgelements for 2.7. I maintain a lot of solid tests for the project, which made tracking down the issues quite rapid. This will be solid python 3 python 2 code. I am temporarily supporting 2.7 again, for bCNC.

tatarize:
See svgelements project for the confirmation of 2.7 support.
https://github.com/meerk40t/svgelements 1.4.3 has support for 2.7 again (briefly).

Harvie:
Thank you. This looks very nice :)

PR #1519, opened 2021-01-15
Z fast before Z enter - Driller

MARIOBASZ:
Added Z order rapid enter, descend at fast speed until surface = surface + 1.00
Added to cut, drill command.
In the driller plugin modified to start in the second movement (the first movement corresponds to the completion of the previous block) that generates drilling twice in the same place.
I do not have a laser, check operation.
In my version of pc, I also generate holes; If these modifications deserve to be merged, and they think it is useful, I will add it in the next intervention


bosd:
Is this still a WIP?

MARIOBASZ:
Wow, it's been two years since I last tried to upload an upgrade!
I'll try to update my github first and then suggest adding fast Z move up to 1mm before contact surface. My machine works like this, jobs become more agile.

Harvie:
Sorry, but i don't like the idea of hardocoding cuting operation with 3 different speeds (rapid, slow feed, fast feed,...).
If you have sandwich material with layers with each having different hardness, you should split your workflow into two separate cuts, each with different speed. If there is someone who wants to cut 5 or 7 layer material? Are we gonna make special case from that as well? I don't think so. Just do two cuts.

PR #1533, opened 2021-02-09
adding experimental pockets with inside islands

DodoLaSaumure:
Hi all

I have added an experimental pocket with internal Islands feature. It is located in CAM/Experimental.
It is written as a plugin, not affecting previous pocket function. At the moment it is still experimental, so I prefer not to interfere with the original code.
In most cases it works fine with simple patterns made of line and arcs. With complex splines, I have detected some issues.
There could still be some bugs, probably because split2contour function inverts some arcs (I suppose, but not sure)
Pierre


DodoLaSaumure:
image

Harvie:
WOW. this is really cool. I am looking forward to test this. I've been working on this some time ago, but i got busy.
BTW what happens when island(s) cut single pocket into multiple separate pockets? This was most prominent edge case that needed to be solved in my implementation... Did you managed to handle this issue?

DodoLaSaumure:
Hi @Harvie 
Do you mean something like this ?
image
I have detected some issues, but it is a promising start.
Pierre

DodoLaSaumure:
or this ?
image

Harvie:
Nice work. We need to figure out how to eliminate unnecessary rapids.

Also... Do you think it would be possible to have "IMAGE" mode which will only mill area laying inside odd number of perimeters? Something like this:

image

image

(basicaly the same stuff all automagic SVG to G-code sign making CAMs do)

Harvie:
Also can you please add brief comments summarizing modifications and newly added functions in bpath.py, what they do and how they compare to preexisting methods? There seem to be bit of overlap with what was already there..

DodoLaSaumure:
@Harvie 

> Also can you please add brief comments summarizing modifications and newly added functions in bpath.py,

This has been done. I have also taken into account the path direction, before making the offset, in order to offset the right direction.
I have tested multiples patterns, under py2 & py3. Seems to work fine.
Still some problem occur with complex splines patterns.

> We need to figure out how to eliminate unnecessary rapids.

Yes, it could be done in a second step.

> Do you think it would be possible to have "IMAGE" mode which will only mill area laying inside odd number of perimeters? 

Do you mean, 
- taking a black and white image as a paramter (jpg, png, etc;.) ?
- or taking lines and filling the odd areas from paths ? (from dxf, etc..) ?

Yes I think it is possible, maybe we can see this in a second step.

All the best
Pierre

DodoLaSaumure:
> There seem to be bit of overlap with what was already there.. 

I do not think they overlap previous functions.
They are slightly different.
For example, 
```
Segment.equals(self,other)
Path.hasSeg(self,s)
Path.isidentical(self,other)
Path.isOnPath(self,P)
Path.isSegInside(self,seg)
Path.isPathInside(self,other)
```
These functions did not exist before.
All the best
Pierre

Harvie:
>    taking a black and white image as a paramter (jpg, png, etc;.) ?
>    or taking lines and filling the odd areas from paths ? (from dxf, etc..) ?

Both :) But don't worry about bitmaps. Everyone can easily trace bitmap in inkscape. So we will have DXF or SVG with outlines of the shapes from the original PNG. Then we have to mill only areas nested inside odd number of outlines. That way the image will get cut correctly.

> I do not think they overlap previous functions.

i don't really checked thoroughly, but some of the functions seemed bit familiar, eg.:

```python
def inside(self, P):
vs.
def isInside(self, P):
```

even if these do slightly different thing, the naming is greatly confusing.

Please note that `bpath.py` is probably the most important file in bCNC (at least as far as CAM goes) and its clarity is absolutely crucial for new developers to be able to produce new interresting plugins easily (like the one of yours). Would you mind renaming your new functions to be more clear? eg. `isInside()` to `isInsidePath()`? I would even rename `inside()` to `insideSeg()` but this would need refactoring of whole codebase, which i don't want to do in this PR to keep things small and easy to review.

Harvie:
I also like the idea of internal pocket feature of bCNC being replaced by plugin (i want to remove the internal stuff and break it out to plugins). Also i would really love to have "profile" feature being just special case of pocket which goes only 1 offset deep.

DodoLaSaumure:
Hi Harvie

> i don't really checked thoroughly, but some of the functions seemed bit familiar, eg.:
> Would you mind renaming your new functions to be more clear? eg. isInside() to isInsidePath()? I would even rename inside() to insideSeg() but this would need refactoring of whole codebase, which i don't want to do in this PR to keep things small and easy to review

```python
def isInside(self, P):
```
was an existing function of bpath.py, 
It returns if a point is inside a path.
I did not want to rename it to avoid confusing people who have their habits, like you, and did not want to refactor the whole code.

unless I do not see something, I only added the following functions
```python
Path.hasSeg(self,s)
Path.isidentical(self,other)
Path.isOnPath(self,P)
Path.isSegInside(self,seg)
Path.isPathInside(self,other)
```
trying to give explicit names.

I carefully read my addtions [in my last commits](https://github.com/vlachoudis/bCNC/pull/1533/commits/398219c0716bd23db0449c5c4a2666f853af7f92) and do not see confusing names. But maybe I have missed something.


> Both :) But don't worry about bitmaps. Everyone can easily trace bitmap in inkscape. So we will have DXF or SVG with outlines of the shapes from the original PNG. Then we have to mill only areas nested inside odd number of outlines. That way the image will get cut correctly.

Seems an interesting idea. Please note that plenty of libraries dealing with images can be found or are already integrated inside bCNC (tkinter could do the stuff, or cv if needed)

All the best
Pierre

Harvie:
Thank you. I am looking forward for more great additions from your kitchen :)

DodoLaSaumure:
> Thank you. I am looking forward for more great additions from your kitchen :)

Thank you Tomas, but I do not work in the noisy kitchen,  rather in my quiet room instead ;-)

Next steps I would like to achieve are :
- Linking pocket paths with G1 moves inside material, instead of G0 moves outside, in this plugin
- Working on the image stuff you described above with odd and even areas, or filling black areas with pockets inside an image, it sounds interesting
- I would like to make a slicer of stl objects, something like [SebKuzminsky/pycam](https://github.com/SebKuzminsky/pycam) does
- and a few small things

Harvie:
> I would like to make a slicer of stl objects.

There is already STL slicer plugin in bCNC. But there is no 3D offset algorithm to make it usefull for more than manual 2.5D processing.

DodoLaSaumure:
Yes I have seen the slice mesh tool.
What  I would like to do is :
- making rough slices from a stl with a cylindrical end mill, to remove the major part of material outside the shape.
- then making a finish pass on the contours of the shape with a spherical end mill

There is another feature that a friend asked me :
It would be nice if we could move the gantry when paused, in order to be able to change a tool during the processing.

Harvie:
> It would be nice if we could move the gantry when paused, in order to be able to change a tool during the processing.

Yes. But it is rather complex issue due to GRBL protocl limitations. As simple workaround you can add rapid to desired position before your toolchange g-code.

PR #1534, opened 2021-02-10
linking pocket paths with g1 moves; split2contours did the job for me ;)

DodoLaSaumure:
Hi @Harvie , I have added the G1 moves to link the pocket paths instead of useless G0 outside of the material
I just added the new segments to the full path, and then reordered all this with split2contour.
It is not optimal in terms of procesing time, we could optimize later to have the path directly ordered, but we have to admit than we can wait some processing seconds when doing a pocket, most of the time is spent on machining the piece of metal on the CNC ;-)

It works fine for me in most cases that I have tested. However, I have found one case where I get strange moves inside the profile, with strange g2 moves that I cannot understand at the moment.

But seems to work fine in most cases
Have a good day
Pierre

### DIFF
--- a/bCNC/plugins/pocketwithislands.py
+++ b/bCNC/plugins/pocketwithislands.py
@@ -71,14 +71,19 @@ def pocket(blocks, diameter, stepover, name,gcode,items):
 				path.name = Block.operationName(path.name, "pocket,conventional,cw", remove)
 			else:
 				path.name = Block.operationName(path.name, name, remove)
-
 			MyPocket = PocketIsland([path],diameter,stepover,0,islandslist)
-			newpath =  MyPocket.getfullpath()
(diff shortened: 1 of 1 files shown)


Harvie:
Yes. Rapid much better now. Thanks!

I assume it would be too bold of me to draw your attention to this issue :-) https://github.com/vlachoudis/bCNC/issues/815
It is related to your plugin and it would help a lot with metal machining to have this...

DodoLaSaumure:
> I assume it would be too bold of me to draw your attention to this issue :-) #815
It is related to your plugin and it would help a lot with metal machining to have this...

Wow there has been a lot of work done around this.

I am going to keep this in mind and will try to help on this if i get an idea.

All the best
Pierre

PR #1535, opened 2021-02-10
allow disable g1 moves inside pockets (to cover strange side effects of split2contour)

DodoLaSaumure:
Hi @Harvie 
As discussed in #1534

> I have found one case where I get strange moves inside the profile, with strange g2 moves that I cannot understand at the moment.

This comes from weird side effects in split2contours

To allow user to generate a pocket in these rare cases, _**I have added the option of disabling G1 segments**_ linking pocket paths.
This is more flexible, and allows the user finding an alternative solution if this occurs, until we have fully understood the problem in split2contours, but I would not like to touch this function at the moment.

All the best
Pierre

### DIFF
--- a/bCNC/plugins/pocketwithislands.py
+++ b/bCNC/plugins/pocketwithislands.py
@@ -34,13 +34,12 @@ except ImportError:
 # =============================================================================
 #==============================================================================
 
-def pocket(blocks, diameter, stepover, name,gcode,items):
+def pocket(blocks, allowG1, diameter, stepover, name,gcode,items):
 	undoinfo = []
 	msg = ""
 	newblocks = []
 
 	islandslist = []
-
 	for bid,block in enumerate(gcode.blocks):
 		if block.operationTest('island'):
 			for islandPath in gcode.toPath(bid):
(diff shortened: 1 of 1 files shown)


Harvie:
Are you aware of "editor->optimize" function provided in bCNC? If you have multiple blocks, it can reorder them to limit excessive rapids between them. Maybe that code can be reused?

DodoLaSaumure:
Hi Harvie

Yes I know the optimize function provided by bCNC and often use it, it's very helpful in many cases.

In this Pull Request, i think it is not usable for this purposes, mainly because 
```python
GCode.optimize(self, items):
```
applies to blocks, and not to paths, or paths lists.

I have read 'optimize()' code, it's very interesting, and maybe we could use a similar algorithm (not sure it applies here however)
optimize functions makes a matrix of the distances between the end points and the start points of blocks, then it orders the block to get the smaller distance.
Not sure the same algorithm can be applied there.

BUT 
There is an issue in split2contours, and we should try to fix this bug.
At the moment, I have not yet analyzed where the bug occurs, so it's to soon for me to fix it.

I just offer a temporary solution for users getting annoyed by split2contours function in the pocketwithislands plugin.
The patch only applies to this plugin and have no side effect on the rest of the code.
It's a temporary solution to help users circumvent the problem.

User has the choice of disabling rapid moves optimization or not. 
Default choice is activate optimization, but if something wrong occurs, you can retry with option disabled.

So I strongly recommend to offer the user the choice until the bug is fixed.

All the best
Pierre

Harvie:
Thanks.
If there is a bug in split2contours, i think everybody will benefit from the fix, because it is used for both DXF and SVG imports and probably in several other CAM operations.

Do you have any example file that i can use to reproduce the bug?

DodoLaSaumure:
Thank you Tomas
Yes i will send you an example file this evening when I have access to my computer

Issue #1536, opened 2021-02-11
Bug in class PocketIsland


DodoLaSaumure:
As discussed in #1535 with @Harvie I am afraid I have discovered a bug in
```python
Path.split2contours(self, acc=EPSV)
```
even if I am not sure yet.

When I make a pocket with inside island on file triangle4.dxf herewith (zipped to allow import inside github comment), I get weird result when I activate the option 
`allow G1 linking segments(default yes)`

When disabled, weird segments disappear.
I use a diam 2 or diam 5 endmill to reproduce.
Here's the file where some problems occur
[triangle4.zip](https://github.com/vlachoudis/bCNC/files/5966840/triangle4.zip)

- With option enabled I get this result with a diam 5 endmill
![image](https://user-images.githubusercontent.com/58803300/107668595-2282fe80-6c91-11eb-8c55-5eb8f38bd82b.png)

- With option disabled I get this
![image](https://user-images.githubusercontent.com/58803300/107668729-45151780-6c91-11eb-877d-bb2f465bea32.png)

I think that the bug occurs when calling split2contous in file pocketwithislands.py, lines 75 -81
```python
			#concatenate newpath in a single list and split2contours
			if allowG1 :
				MyFullPath = Path("Pocket")
				for path in newpathList :
					for seg in path:
						MyFullPath.append(seg)
				newpathList = MyFullPath.split2contours()
```

The bug is not in appending segs in MyFullPath,
```python 
				for path in newpathList :
					for seg in path:
						MyFullPath.append(seg)
```
because I just append Segment.LINE types and we've got an arc in picture 1.

Here are other files where some problems occur with pockets, **both with the old pocket function** and the **new pocketwithislands plugin**.
[horloge.zip](https://github.com/vlachoudis/bCNC/files/5966995/horloge.zip)

- Making a pocket with the standard pocket tool on the C of Besançon provides this, with an diam 1mm endmill
![image](https://user-images.githubusercontent.com/58803300/107669307-e308e200-6c91-11eb-8895-2c580f967b1e.png)

and there are other problems on some letters, with both tools.

In addition, if this can help, 
```bash
grep -r split2contours 
```
gives the following results

```python
lib/bpath.py:	def split2contours(self, acc=EPSV):
lib/bpath.py:			opath = opath.split2contours()
lib/bpath.py:	# FIXME: This probably can be replaced with split2contours() and i've just reinvented wheel LOL

[...]

PR #1541, opened 2021-02-14
Pocket islands plugin

DodoLaSaumure:
try to merge profile, pocket, pocket with islands into a single experimental plugin

### DIFF
--- a/bCNC/plugins/pocketwithislands.py
+++ b/bCNC/plugins/pocketwithislands.py
@@ -39,7 +39,6 @@ def pocket(blocks, RecursiveDepth,ProfileDir,CutDir,OverCut, CustomRecursiveDept
 	undoinfo = []
 	msg = ""
 	newblocks = []
-	print ( " RecursiveDepth,ProfileDir,CutDir,OverCut,",RecursiveDepth,ProfileDir,CutDir,OverCut)
 	islandslist = []
 	for bid,block in enumerate(gcode.blocks):
 		if block.operationTest('island'):
--- a/bCNC/plugins/pocketwithislands.py
+++ b/bCNC/plugins/pocketwithislands.py
@@ -34,11 +34,12 @@ except ImportError:
 # =============================================================================
 #==============================================================================
 
-def pocket(blocks, allowG1, diameter, stepover, name,gcode,items):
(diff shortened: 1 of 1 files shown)


DodoLaSaumure:
@Harvie 
I have not added yet the extra segments to cover 100% of the pocket in all cases (uneven lines problem with stepover > 50%)
But I think it covers all the other points that we reviewed in #1536.
I have done some tests, for me it appears to be ready to be merged in head. 
Please make some further tests if you want to be sure.
Pierre

Harvie:
There is bug in offset direction and overcuts.

This is original profile tool in bCNC (inside+overcut):

image

This is your plugin (inside+overcut=1):

image

Your plugin (outside+overcut=1):

image

So you can see the direction is inverted and overcuts do not work.
Please make overcut as checkbox rather than numeric input. (is there any benefit from this?)

Also see how offset direction (inside/outside) works in original code. We had similar problem in there. If i remember corretctly the solution was to normalize the path direction (CW/CCW) to be same in all cases. Then it was easy to get consistent switching between inside/outside profiles.

Harvie:
OOOH. I understand now.
You've confused "overcuts" and "additional offset distance". This is what seems to cause the issues.

Overcuts are these tiny pieces of path that allow to cut all the way into the corners so you can avoid corners being rounded and fit square peg into the square hole:

image

It is very important!

image

DodoLaSaumure:
> You've confused "overcuts" and "additional offset distance".

ah ok, yes I confused these two things.
Going to fix it soon
Pierre

DodoLaSaumure:
In this case, it is not exactly an additional offset distance (it is substracted from the offset), it is a correction for correcting the path tolerances (like a tool correction)
it's very important too, to be able to adjust the tolerance of a profile.

DodoLaSaumure:
I would name it "additional distance inside profile", I think that's more clear to all.
Pierre

Harvie:
Thank you... So far i was not able to find any issue.
Only thing that comes to my mind is ignoring the islands:

1.) Add checkbox "ignore islands", so users can easily get back the original behaviour
2.) By default ignore all islands marked as "tab"... I see very little point in affecting profiles and pockets by tabs. These will be handled in cutting operation anyway...

PR #1543, opened 2021-02-15
overcuts and additional cuts

DodoLaSaumure:
Add overcuts and additional cuts inside corners

### DIFF
--- a/bCNC/lib/bpath.py
+++ b/bCNC/lib/bpath.py
@@ -1433,6 +1433,7 @@ class Path(list):
 	# a certain distance
 	#----------------------------------------------------------------------
 	def overcut(self, offset):
+		overcuts= Path("overcuts")
 		if self.isClosed():
 			prev = self[-1]
 			Op = prev.orthogonalEnd()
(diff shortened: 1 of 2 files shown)


DodoLaSaumure:
And you can now manage overcuts inside / outside pockets :)

DodoLaSaumure:
Maybe we can reduce rapid moves between overcuts, but as I am not sure to succeed, I prefer pushing this previously to attempt

DodoLaSaumure:
That was easy in fact. ;)

DodoLaSaumure:
I think it's ok now.
Pierre

Harvie:
Thank you... So far i was not able to find any issue.
Only thing that comes to my mind is ignoring the islands:

1.) Add checkbox "ignore islands", so users can easily get back the original behaviour
2.) By default ignore all islands marked as "tab"... I see very little point in affecting profiles and pockets by tabs. These will be handled in cutting operation anyway... So perhaps tabs should be ignored always.

Harvie:
Also it might be useful to add option "only regard selected islands", so user can select which islands to use.

When this is done, i will try to integrate this to the bCNC UI.

DodoLaSaumure:
Hi Thomas, 
This has been done in #1545

PR #1545, opened 2021-02-15
add ignore islands option  / add some doc help

DodoLaSaumure:
Hi

- I have added ignore islands option => choice among the following list :
Regard all islands except tabs,
Ignore all islands,
Regard only selected islands

- I have added some help inside the plugin

### DIFF
--- a/bCNC/CNC.py
+++ b/bCNC/CNC.py
@@ -4106,7 +4106,7 @@ class GCode:
 		blocks.extend(newblocks)
 		return msg
 
-    #----------------------------------------------------------------------
+	#----------------------------------------------------------------------
 	# make a trochoidal profile on block
 	# offset +/- defines direction = tool/2
 	# return new blocks inside the blocks list
--- a/bCNC/plugins/pocketwithislands.py
+++ b/bCNC/plugins/pocketwithislands.py
@@ -34,20 +34,37 @@ except ImportError:
 # =============================================================================
 #==============================================================================
 
-def pocket(blocks, RecursiveDepth,ProfileDir,CutDir,AdditionalCut,Overcuts, CustomRecursiveDepth,
-		allowG1, diameter, stepover, name,gcode,items):
+def pocket(selectedblocks, RecursiveDepth,ProfileDir,CutDir,AdditionalCut,Overcuts, CustomRecursiveDepth,
+		ignoreIslands,
+		allowG1, diameter, stepover, name,gcode):
 	undoinfo = []
 	msg = ""
 	newblocks = []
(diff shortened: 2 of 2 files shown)


Harvie:
Great work. Thank you.

I've already released latest version of your plugin so it can be installed using pip...

I will try to figure out, how to integrate it in the bCNC... So far i've done some tests and i know how to change original Profile and Pocket buttons in bCNC to open PocketIsland plugin instead... BUT i would like to figure out how to preconfigure the plugin for doing pocket or profile based on which button user uses... Maybe i will need to restructure the GUI even further.

If you have any ideas how to get rid of these missing lines when using higher stepover, please go try to figure it out, because it triggers my OCD and is kinda suboptimal:

image

DodoLaSaumure:
Thank you Tomas.

> If you have any ideas how to get rid of these missing lines when using higher stepover,

Yes I have an idea, but I am afraid of side effects.
going to make some tries..

DodoLaSaumure:
in the picture you show, it is normal to have a missing line in the middle of the shape : 
The two center lines are separated with a bit diameter, while the other lines are separated by a half bit diameter.
This is normal, 100% of the material is removed inside the pocket in this case.

Where it sometimes is an issue is when an island intersects an outpath with an uneven line.
But adding the missing line can cause other problems, I am not sure
I will make some tries.

Also I have found a bug with the overcuts, in one very rare case, with islands inside a the shape.
I have an example that reproduces the bug.
I am going to investigate....

DodoLaSaumure:
> in the picture you show, it is normal to have a missing line in the middle of the shape :
The two center lines are separated with a bit diameter, while the other lines are separated by a half bit diameter.
This is normal, 100% of the material is removed inside the pocket in this case.

In fact there should be no more than 1 * diameter between middle lines and no more than 1* diameter * stepover% between other.
When stepover > 50% and when an inside path intersects an out path with a small angle there can remain uncut area that Can be resolved by setting stepover <= 50.

The  idéal solution is to link consecutive intersections with the appropriate segment which is not so easy...

Harvie:
>  This is normal, 100% of the material is removed inside the pocket in this case.

This is not true. We are not using 100% (or even 95%) stepover for a reason. It might work in CNC simulator. But in reality this will cause troubles. Especialy when milling flexible materials like some kinds of aluminum or most of the plastics with high stepover. There will be thin wall of material between the cuts that will bend away from the milling bit and will not get cut.

DodoLaSaumure:
Understood. I will have a look soon.
All the best
Pierre

PR #1549, opened 2021-02-17
Uneven lines this is a fix for issue found bye @Harvie in #1545

DodoLaSaumure:
Hi Tomas, 
I think I have found why the reason why last pass was sometimes uneven.
What I am doing is 
- detect we are in the last pass
- if yes, instead of applying standard offset, apply min(diam*stepover/2, diam*0.49)

This should solve the problem detected in #1545

### DIFF
--- a/bCNC/plugins/pocketwithislands.py
+++ b/bCNC/plugins/pocketwithislands.py
@@ -160,8 +160,10 @@ class PocketIsland:
 	def inoutprofile(self):
 		if self.depth == 0:
 			self.offset = -self.diameter / 2.0 +self.AdditionalCut
+			self.offsetLastPass = self.offset
 		else:
 			self.offset = -self.diameter*self.stepover
+			self.offsetLastPass = -min(self.diameter*self.stepover/2.,self.diameter*0.49)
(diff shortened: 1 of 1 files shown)


Harvie:
Thank you.

This is slightly better. I use 40% stepover and it does 60% in some cases, while not perfect i think i can live with that. Overall it seems to work better than previous plugin version.

What is buging me now is the amount of G0 moves, because my machine has relatively slow Z axis, so doing Z hops inside pocket makes everything quite slow.

DodoLaSaumure:
> What is buging me now is the amount of G0 moves,

Did you consider enabling option **allow pocket paths linking segments(default yes)**
If yes, I do not see one more can be done.
Pierre

Harvie:
> Did you consider enabling option allow pocket paths linking segments(default yes)

Yes. Sometimes it adds weird link to nonexistent (deleted?) path:

image

> If yes, I do not see one more can be done.

I beleive the linking can be reordered and optimized to use lot less rapids. For example i am sure this whole pocket can be cut completely without rapids, because there are not gaps between toolpaths:

image

PR #1553, opened 2021-02-23
status message progression

DodoLaSaumure:
Hi Tomas, I have added some messages in the status bar, to show progression for long pockets.
All the best
Pierre
(Tested under py2, py3, py3.9)


Harvie:
Nice. Thank you.
BTW what do you think about adaptive clearing strategy? There is python code from FreeCAD for that:
https://github.com/vlachoudis/bCNC/issues/815#issuecomment-524931273
But i have not yet checked how hard would it be to modify for use with bCNC.

Having this would greatly increase chances of machining aluminum using bCNC and hobby machines. As well as it would enable more effective machining of wood.

DodoLaSaumure:
@Harvie 

Thank you for all your efforts for maintaining bCNC.

> BTW what do you think about adaptive clearing strategy? 

At the moment I have no idea about this, please give me some time to think about it.
My idea is that yes we could use FreeCAD code, instead of reinventing the wheel.
And I do not think that endmillofffset.py can easily be reused.
But I have to take the time to think about this.

Also I have tried to use slicemesh, and found plenty of memory leaks (each time I load a binary stl, the system goes out of memory after several minutes). I think it would be great if we could simplify the code to a minimal stl file loader, and generate 2 cuts
- One Cut with a cylindrical mill inside a rectangular contour to do the approach with different strategy cuts ( along x, along y, or along stl)
- One cut for the finish with a ball end mill, 

As I have read your stl slice mesh code, I would not like to interfere with your works, but I would like to do a much more simple architecture inside an independent plugin.


What's your opinion about this, please ?

All the best
Pierre.

Harvie:
> At the moment I have no idea about this, please give me some time to think about it.

No stress, i am aware that this is not an easy task :-)

Another interresting aproach is the circular/trochoidal milling strategy... Which is almost OK in bCNC plugin i've made, but the tabs are really stupid right now, causing lot of plunging, which is not good in such kind of operation.

> As I have read your stl slice mesh code, I would not like to interfere with your works, but I would like to do a much more simple architecture inside an independent plugin.

Totaly. I have created this plugin for real-life use case, where customer provided me with 2.5D STL model which i've sliced and processed using 2D operations available in bCNC. So i think it is already useful as is and i think the functionality should be preserved. So for me it makes sense to start as separate plugin.

Harvie:
I had some issues with CAMotics g-code simulator. Now i've fixed it and realized terrible thing.
When i do Pocket using your code and then Cut it in several Z passes. It goes DEPTH FIRST on several parts of the path.
That is bad and hugely increases chances of breaking bits.

Best solution for this would be to figure out, how to do G1 linking between ALL segments in the pocket. (obviously with exception of pocket being split to multiple separate pockets by island).

Do you have any idea on how we can add missing G1 links?

Harvie:
For example, i have pocket like this:

image

But it first cuts some parts of the pocket all the way down, which is wrong because causes unnecesary deep grooving which prevents proper chip evacuation:

image

Harvie:
We probably need to find closest two points of these individual paths (or create such point by breaking segment). To find proper g1 links.

Harvie:
For example the highlighted part is not linked:

image

DodoLaSaumure:
Yes I am aware of this, but I have no better solution until now.

And I am not sure there is one
because
image

- The first two offsets are linked towards the lower left 
- The  last offsets are linked too

So after linking all these path, you'll have to go Z+ to jump towards the remaining blue path.
I am not sure there is a way to avoid this, unless going several times on the same paths.
Pierre

Harvie:
> unless going several times on the same paths.

yes, this is obvious solution and i don't think that there is something wrong with such solution. it's much better than recutting chips deep down in the groove.

DodoLaSaumure:
>>    unless going several times on the same paths.

> yes, this is obvious solution and i don't think that there is something wrong with such solution. it's much better than recutting chips deep down in the groove.

I have not been using a professional CAM software since 1995, I only use bCNC for my personal hobbies, and no other software.
I would be curious to know how they proceed pockets and link paths inside pockets, if someone has the answer.

going several times in the same paths seems to be very complicated in the recursive architecture of the PocketIsland class. 
And regenerating these segments after the full recursion, seems to be very complicated, too....

Harvie:
> I would be curious to know how they proceed pockets and link paths inside pockets,

2D adaptive toolpath with spiral plunge. But that would need lot of work in bCNC.

Harvie:
Btw just found there is this Adaptive Clearing Tool Path Library
https://github.com/jf---/libactp
https://github.com/Heeks/libactp-old

PR #1555, opened 2021-02-25
Fix unneeded executable file permissions

MicroJoe:
Some files are marked as executable, but do not need to.
Here are the files that are fixed:

* locale: fix *.mo and *.po perm to 0644
* bCNC.desktop: fix perm to 0644


Harvie:
Thanks

PR #1556, opened 2021-02-25
use upstream svgelements instead of vendoring

MicroJoe:
It is preferred to use pypi instead of copying code of other projects, unless justified. This was tested to be working by importing a simple .svg file into bCNC and it worked without any issue.


Harvie:
I beleive there was some reason for this...

MicroJoe:
> I beleive there was some reason for this...

Do you remember? Maybe because svgelements had some issues, but it seems to be well maintained and has fixed some bugs in the meantime. The manual import is from 2018 IIRC, but svgelements last release is from 2020. How can I test that this will not cause any issues?

Harvie:
There was some requirement by @vlachoudis (original author) that we use as little external dependencies as possible.
But hey, that was before i've introduced installation using pip, so i guess it makes more sense now to use external deps.

Harvie:
Can you please rebase this to resolve the conflict?

tatarize:
If this is to be merged, #1720 is unneeded.

tatarize:
The vendoring was due to the py2/3 requirement. The library was not py2 compliant after eol. But, after #1714 was merged that is no longer needed.

The conflict is due to changes in svgelements from things like the aforementioned #1714, but the file is being deleted. Deleted is the correct merge there. I checked the glue code again, it all looks fine. None of that has changed. Though there could be reasons to opt for interpolating the bezier curves as circular arcs. But that could be handled in a different PR. In any event you just make circular arcs with control points that match points on the bezier curve and make sure the divide the curve at the inversion points (so things don't get weirdly messy.

Harvie:
>  Though there could be reasons to opt for interpolating the bezier curves as circular arcs.

i don't really recall that well, but i beleive there was some bezier interpolation right? is this some local modification that we will loose by using upstream version of library?

MicroJoe:
> Can you please rebase this to resolve the conflict?

Done

Harvie:
Thanks.
Hope it did not broken bezier import :-)

tatarize:
You wouldn't actually lose it, and it's currently not done. But gcode, uses two segment types, these are lines and circular arcs. There's a couple weird G5 G6 things that I don't think anybody really supports but generally all the stuff supports G3, G4. Now, for the bezier curves we're actually interpolating from the svg. So we take like 50 points along the curve and make that out of lines. However, there's also interpolation via circular arcs that could be done. Generally this is done by taking a number of segments finding 3 points along the curve and just drawing a circular arc through those three points.

The current glue code, before or after merging this did the linearization there.

The newer version of this same code actually has an additional bit of code:

```python
    def approximate_bezier_with_circular_arcs(self, error=0.01):
        """
        Iterates through this path and replaces any bezier curves with circular arcs.
        """
        for s in range(len(self) - 1, -1, -1):
            segment = self[s]
            if isinstance(segment, (QuadraticBezier, CubicBezier)):
                self[s : s + 1] = list(segment.as_circular_arcs(error=error))
```

Which wasn't an option before but could be an option here after using the more modern library. You should not lose anything, the modern stuff is a bit better across the board and doesn't do anything weird that I know of, it's mostly been bug fixes for a couple years now. 

Previously this circular arc interpolation wasn't a thing:

```python
    def as_circular_arcs(self, error=ERROR, start=0.0, end=1.0):
        mid = 0.5 * (end - start) + start
        start_point = self.point(start)
        mid_point = self.point(mid)
        end_point = self.point(end)
        sample_point0 = self.point(0.25 * (end - start) + start)
        sample_point1 = self.point(0.75 * (end - start) + start)
        arc = Arc(start=start_point, end=end_point, control=mid_point)
        radius = Point.distance(start_point, arc.center)
        r0 = Point.distance(sample_point0, arc.center)
        r1 = Point.distance(sample_point1, arc.center)
        current_error = (abs(r0 - radius) + abs(r1 - radius)) / radius
        if current_error < error:
            yield arc
        else:
            yield from self.as_circular_arcs(error=error, start=start, end=mid)
            yield from self.as_circular_arcs(error=error, start=mid, end=end)
```

See:
https://pomax.github.io/bezierinfo/#arcapproximation

For a description of how this is done. Note, this was not done in the original code but *is* an option given that gcode should have support for circular arcs. 

This is the current glue-code:

```python
            if isinstance(segment, Move):
                gcode.append(f"G0 X{rv(segment.end.x)} Y{rv(-segment.end.y)}")
            elif isinstance(segment, (Line, Close)):
                gcode.append(f"G1 X{rv(segment.end.x)} Y{rv(-segment.end.y)}")
            elif (isinstance(segment, Arc)
                  and abs(segment.rx - segment.ry) < 1e-9):
                # Strictly speaking, svg arcs can be non circular,
                # whereas gcode only permits circular arcs.
                garc = "G02" if segment.sweep > 0 else "G03"
                gcode.append(" ".join([
                    f"{garc}", f"X{rv(segment.end.x)}",
                    f"Y{rv(-segment.end.y)}", f"R{rv(segment.rx)}"
                ]))
            else:  # Non-circular arc, Cubic or Quad Bezier Curves.
                subdiv_points = numpy.linspace(0, 1, subdiv, endpoint=True)[1:]
                # numpy accelerated point() call
                points = segment.npoint(subdiv_points)
                gcode.extend(
                    [f"G1 X{rv(sp[0])} Y{rv(-sp[1])}" for sp in points])
```

Basically G02 and G03 are sent if you have an arc segment where rx == ry. Otherwise it cuts it up into subdiv number of points and draws lines between each such point. Though really, in theory, a call to `approximate_bezier_with_circular_arc` there could let you write that curve as a series of arcs. It doesn't say but it's actually in the curve class and would apply to arcs too. So you could certainly approximate the non-circular arcs with circular arcs.

We cycle through and let you replace the non-circular arcs curves in the path and write those out. 

I brought it up not because it would be lost, but because it's an interesting concept, though it might belong more obviously in bCNC rather than just in the SVG import code. Heck you could include support for G5 and maybe interpret them later to arcs or lines at your discretion. 

https://observablehq.com/@awhitty/approximating-bezier-curves-for-cnc

Basically SVG has non-circular arcs, quads, and cubics these (other than quads under very limited use G05) aren't supported by gcode and need to be made into either lines or circular arcs. Either one of those is actually an option.

tatarize:
Sounds roughly able to do that. Though here we *already* have some bezier curves, so converting them to tiny lines and then converting that to arcs might be a bit lossy, but but unless there's somehow G5 Quad Bezier commands permitted in the code I'm pretty sure  such things will likely take a minor loss and linearization is pretty reasonable (and I think like Marlin converts the G5 to 10 line segment moves in their controller anyway rather than using zingl-bresenham or some of the fancier solutions there.

Harvie:
Only lines and circular arcs are allowed in bCNC. While grbl might be able to process splines and bCNC will happily stream them if you have g-code for them, it would be extremely painful to add spline support to bCNC in terms of any post-processing or even CAM, because now basicaly every bCNC feature has separate code for lines and arcs and it's already complex.

Basicaly any CAM feature of bCNC has to handle all segment types separately. Eg.: imagine offset/pocket operation. you need separate code to make line offset and separate code to make arc offset. then you need code to properly connect intersection of the two segments together in corners (eg. shorten/lengthen them to perfect trim). such code has to be done for all possible combinations of segment types. (line-line, line-arc, arc-arc). So basicaly these are special cases all the way. Now imagine adding new type of segment, that would make everything so complex it would be probably really hard to sustain the development with resources available to bCNC. and would dramaticaly increase cost of adding new CAM features to bCNC.

In my experience the ArcFit can give pretty good results when configured correctly, i was fiting complex splines with it and it did reduced numeber of g-code lines by several orders of magnitude without sacrificing aesthetics of processed shape in any noticeable way.

PR #1557, opened 2021-02-25
remove removed lib/svg library

MicroJoe:
This library is not present anymore but is still mentionned in some
places.

### DIFF
--- a/LICENSE.MIT
+++ b/LICENSE.MIT
@@ -6,10 +6,6 @@ Copyright (c) 2015 Julien Rebetez
 	lib/ply.py
 	lib/utils.py
 
-svg.path ( https://github.com/regebro/svg.path )
-Copyright (c) 2013-2014 Lennart Regebro
-	lib/svg/*
-
 svg.elements ( https://github.com/meerk40t/svg.elements )
 Copyright (c) 2019 David Olsen, 2019 meerk40t, Zingl
 	lib/svg_elements.py
--- a/makefile
+++ b/makefile
@@ -4,8 +4,6 @@ SOURCES =	bCNC/*.py \
 		bCNC/lib/*.py \
 		bCNC/lib/python_utils/*.py \
 		bCNC/lib/stl/*.py \
-		bCNC/lib/svg/*.py \
-		bCNC/lib/svg/path/*.py \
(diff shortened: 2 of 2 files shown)


Harvie:
If you are into this stuff, i would prefer some code cleanup of STL and PLY imports.

Namely to strip STL import down to single file. Similar to how PLY is imported.
And to extend PLY import binary format (currently only ASCII is supported).

MicroJoe:
> If you are into this stuff, i would prefer some code cleanup of STL and PLY imports.
> Namely to strip STL import down to single file. Similar to how PLY is imported.
And to extend PLY import binary format (currently only ASCII is supported).

Nice to know, I will try to do this in another pull-request. In the meanwhile please let me know if anything from this one is incorrect. A little bit of cleanup will help new contributors to not get lost in the used dependencies (now using svgelements instead of lib/svg/).

tatarize:
Yeah. `svgelements` is a descendant of `svg.path` from way back but with actual svg parsing rather than just basic path. During the last update after I solved a lot more edge cases and it was stable and had vectorization to make the loading *much* faster, I removed the last bits of `svg.path` since it hadn't been used since the initial switch to `svgelements` . I only checked that no code anywhere referenced it, not that there wasn't any documentation etc.

tatarize:
The older name was svg_elements though really it was svg.elements but the dot was non-pythonic and thus dropped. Version "1.4.2" could included with pypi since I briefly went back through and ported the stuff over to python 2 for the last round of upgrade, and included that in the pypi manifest etc. I only included it directly since it seemed the way things were being done.

Harvie:
> This library is not present anymore but is still mentionned in some

BTW you are quite thorough... do you use some automatic tool to find unnecessary libraries in the project?

MicroJoe:
> > This library is not present anymore but is still mentionned in some
> 
> BTW you are quite thorough... do you use some automatic tool to find unnecessary libraries in the project?

Mhm, not really. I have found some issues using [Lintian](https://lintian.debian.org/), but not this one. Packaging for Debian requires careful reading of the `COPYRIGHT*` files. I found that this part was not required anymore so I took the opportunity to write this small contribution.

PR #1559, opened 2021-02-27
Fix bug in dragknife plugin

zyx:
The dragknife plugin currently produces arcs that fail validation when
sent to grbl.

When the tangentialOffset method is called on an arc segment, both it's
start and end position are offset. The center is moved such that it is
equidistant between these new positions. This produces a valid arc.
However, when turned into GCode, only the center and end positions are
used - the start position is assumed to be the end position of the
previous command. This _isn't_ a valid arc, and GCode processors like
grbl will report errors when sent this GCode.

This commit fixes the problem by using the end of the previous (tangentially
offset) segment as the start of subsequent offset arc segments.

### DIFF
--- a/bCNC/plugins/dragknife.py
+++ b/bCNC/plugins/dragknife.py
@@ -109,7 +109,7 @@ This fact introduces the need for preprocessing the g-code to account with that 
 					else:
 						arcdir = Segment.CCW
 
-					#Append swivel if needed (also always do entry/exit)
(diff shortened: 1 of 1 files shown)


Harvie:
Good point!
I think there might be similar error in the "ArcFit" tool of bCNC... Would you mind checking that as well please?

Harvie:
Thank you.

PR #1561, opened 2021-02-28
Stl 3D slicer

LittlePierre:
This is a first draft of an experimental simple stl binary 3D slicer.
This first draft only accepts valid binary stl files and does not check for errors.

The ball nose mill finish is not yet implemented.
And the Cylindrical nose mill does not check yet if it enters the previous upper slices.

But this is a beginning, starting to work, and can give some ideas to / from others

Pierre


DodoLaSaumure:
For my tests I used the following stl file
[ScaledShoeNOHEEL3.stl.txt](https://github.com/vlachoudis/bCNC/files/6056846/ScaledShoeNOHEEL3.stl.txt)
The parser and slicers appear to be very fast, comparatively with the previous one. (even if not handling all errors yet)

DodoLaSaumure:
To give an idea, i get this
image
or this
image

Harvie:
> And the Cylindrical nose mill does not check yet if it enters the previous upper slices.

I've just tested the code and it does not seem to be more useful for 3d milling than my previous "Slice Mesh" plugin. But i understand that this might need lots of effort and i am looking forward for new development in this area.

I will not merge this *YET*, but that does not prevent me from being closely observing any further progress of yours.

Also please note that bCNC has poor rendering engine. Slooooooow (most of it is python, no optimized C code), no antialiasing, no interactive mouse rotation, perspective not configurable. This is huge handicap making it next to useless for inspecting complex 3D shapes and working with stuff like that. We need to find a way how to implement accelerated 3D g-code canvas running mostly as precompiled binary module.

There was some discussion in #591 , but so far didn't managed to get anything running. Some of the options are limited to python3, which is luckily no longer and issue for bCNC, so we might review the options again. There are python modules that allow to create interactive 3D visualisation UI and i think it would make sense to use some of them. Unless we want to write that stuff from scratch in C/C++...

Personaly i think this absolutely has to be done before we start going near any "actual 3D" processing. Even for 2D the performance of canvas is huge PITA. We can have awesome CAM working with 3D meshes, but if we take 60 seconds to render output of that slicer in sh*tty graphics, nobody will care.

DodoLaSaumure:
> Also please note that bCNC has poor rendering engine

Yes, Tk is not the best 3D rendering engine ;)

But for a 10688 triangles stl
I get the following performances : 
4.74 seconds for 1-Rough Slice rough removal(cylindrical nose) (Using Diam 8mm cylindrical mill, and making 8 slices of 8mm each)
4.63s for 2-Rough Slice finish(cylindrical nose)(Using Diam 8mm cylindrical mill, and making 8 slices of 8mm each)
for full compute and render.
And my PC is an old laptop.

That's much better than the previous slicing tool, and than [PyCam](https://github.com/SebKuzminsky/pycam) does with openGL.... ;)

Harvie:
I never used PyCAM (or BlenderCAM), because it always seemed terribly slow to me. Both are slow because of Python.
Rendering and slicing/processing are two different things. We have reasonably fast CAM. But we have slow canvas.

Harvie:
> Yes, Tk is not the best 3D rendering engine ;)

It is possible to embed OpenGL context in tkinter GUI. Similar way as you would embed it in GTK or Qt GUI... But it gets bit rough around edges and you will not gain much performance without actualy moving the processing of g-code to OpenGL from python to C. Maybe we might get some improvement by converting all g-code to numpy vectors and doing 3D rotations in it. But i don't know how far this can get us.

DodoLaSaumure:
We can now make
- Raw Slices (with tool offset) - For visualization only
image

- Rough removal along direction x or y, preventing from milling inside previous Slices - Cylindrical end mill
image


- Rough Contours, along contours of stl, preventing from milling inside previous slices  - Cylindrical end mill
image

- Not done yet 
Remove material inside internal wholes ( internal contours are for now eliminated)
Finish surface with a ball end mill

DodoLaSaumure:
here's the gcode generated after 17 passes of 3mm ( Rough removal + Rough contour) with a 8mm cylindrical end mill on the previously uploaded [3d model](https://github.com/vlachoudis/bCNC/files/6056846/ScaledShoeNOHEEL3.stl.txt)
[ScaledShoeNOHEEL3_contour.ngc.txt](https://github.com/vlachoudis/bCNC/files/6101367/ScaledShoeNOHEEL3_contour.ngc.txt)
[ScaledShoeNOHEEL3_rough.ngc.txt](https://github.com/vlachoudis/bCNC/files/6101369/ScaledShoeNOHEEL3_rough.ngc.txt)



here's a view from CAMotics, rendering the result 
image

Harvie:
Looks promissing. Any idea how to tackle finishing strategy?
Also if i remember correctly, the roughing pass should leave some extra material (both radialy and axialy) so finishing pass can cut consistently.

Harvie:
Also look at strategies described in here: http://blendercam.blogspot.com/p/blender-cam-description.html

Basic X/Y grid based finishing:

image

Spiral strategy can be interresting addition for some objects:

image

DodoLaSaumure:
Thank you for your intersest and advices Tomas.

> Any idea how to tackle finishing strategy?

In a first step, I would like to make something rather basic.
So for the cutting strategy I am going to implement only along x / along y strategy cuts.
At least in a first step.  And then depending on users needs, and on my availibility to develop further options. Or maybe another developer can add extra features if needeed....
I think it covers most usages.
And anyway finish pass does not prevent from polishing the finished surface. The user has to work with his hands a little ;) 


> Also if i remember correctly, the roughing pass should leave some extra material (both radialy and axialy) so finishing pass can cut consistently.

There is an option "offset", allowing user to add a radial tool correction. (offset positive inside material, to be clarified in the help interface)
For the axial correction : not implemented yet. Seems very easy to do by changing Z of current pass, after path computation.


All the best 
Pierre

DodoLaSaumure:
Notes for myself, and whoever is interested in this algorithm.

After some time away from the project, I finally got some free-time to work on it again.
Here’s some theory about the algorithm that I am trying to implement.

A Current algorithm
1°) the current algorithm does not take into account holes inside the materials (also called islands in my description)
2°) The algorithm makes slices of the stl file, going from the top to the bottom.
3°) Currently, it only takes the outer profile of the previous slice and the current slice, to avoid the mill to go into the freshly machined slice.
4°) The result is stored into the current slice, for the future lower slices.


B Algorithm taking into account the holes
1°) Take a slice, with severals path corresponding to the holes contours (empty material) and plain material contours. You get several paths.
2°) For each path, check if it is plain material contour or hole contour :
a) Take a point inside the contour, draw a segement from the outer point of the working area, and count intersections : if odd, this is plain material, if even, this is a hole.
b) Tag all the paths if plain or empty.
3°) Go to next slice, and process again.

The figure below describes how the plain and empty contours can interact (  previous / current slice)



To parse the slices, take the outer profile of current slice and previous slice, and check how they interact (fig 1-2-3 plain material / plain material)
Then go inside the material and check the outer islands
And so on.

I think that this should cover most cases.
Of course, avoid using this type of plugin for flat surfaces, offset, pocket and contour do the job much better for this




slicerIslandsTheory

 Any advices / Remarks / Comments are welcome
Thanks
Pierre

DodoLaSaumure:
Hi

I have thought to it again.
In fact the problem seems simple but is not so easy, because I do not have a 2D map of the plain and empty areas, but only the contours of the paths, tagged plain or empty.

image


Blue is the previous slice
Red is the current slice.
I want to get the contours of the white areas.
But I do not have a 2D map of the hatched surfaces, this would take to much memory, and would not be accurate to follow the contours.
I only have the paths of the contours, tagged plain or empty.

The process would be something like this :

1) Get the outer part blue profile
Get the outer part of the red profile.

If they intersect, take the outer part of the interection
If not, take the outer path.

Store the path in the new path list, and tag it as plain.


2) Take the outer paths of the red profile inside the new contour ( there can be several)
Take the outer path of the blue profile inside the new contour (there can be several)

For each red :
if it is in a plain area of the blue, ignore
if it is in an empty area of the red, take the inner and store it in the new path list, tag it empty
if red and blue interesect take the inner part of the intersection, and tag it as empty

For each blue, process the same

With each new path added in the the new path list, goto 1 (recursive function)

Any ideas ? Comments ? Remarks ?

Harvie:
There is actualy some code in bCNC for this. It takes arbitrary point inside of the shape and draws a line to near infinity. Then counts how many times this crosses some contour. Then you simply determine if you have crossed odd or even number of contours. I am pretty much sure there is code for this already...

DodoLaSaumure:
Thanks Tomas

Yes I know this code

it is in bpath.py lines 1100 & following

```python

	def isInside(self, P):
		#print "P=",P
		#minx,miny,maxx,maxy = self.bbox()
		maxx,maxy = self.bbox()[2],self.bbox()[3]
		#print "limits:",minx,miny,maxx,maxy
		#FIXME: this is strange. adding +1000 to line endpoint changes the outcome of method
		#	i've found that doing this works around some unknown problem in most cases, but it's not really ideal solution
 		line = Segment(Segment.LINE, P, Vector(maxx*1.1, P[1]+1000))
```


it has to be adapted to the current use case, and we have to parse every segment of every path, one to one, which can take some time for processing. I was wondering which would be the optimal way to proceed, I was just sharing my thoughts ...
But sure it is a good idea to reuse.

DodoLaSaumure:
here we are !!! We have now the internal holes contours milled
I put a commit in a few days
Still to do, 
- ball end mill finish, that should be very easy now (parse along direction x or y, take the z ball-offset the xy slice, take the zmax of the slice)
- Code cleanup

image

image

Harvie:
> I used the "bruteforce method" described above

Would it help to have "minkowski sum" in bCNC library? I've heard that people are using minkowski to efficiently caculate offset toolpath for complex geometry and endmills with arbitrary shapes...

DodoLaSaumure:
> Would it help to have "minkowski sum" method implemented in bCNC library?

I learn from you everyday Tomas. I never heard before of the Minkowski Sum. ;)
I have read some documentation, it could help for endmills with complex shapes. But now the plugin is almost finished, and I plan to implement cylindrical and ball end mills only, at least in a first step. 
So could be useful for future purposes, but not sure I need it now, I am afraid it would confuse me to finish the plugin.

Thanks
All the best
PIerre

DodoLaSaumure:
image

DodoLaSaumure:
Here's an alpha release of ball finish.
Complexity depends on the number of slices intersected by the ball end mill, corrected with the radius offset, and the toolstep
given
* r is the radius of the mill
* offset is the radius offset
* toolStep is the step in the slices
Complexity is ((r+offset)*2+1)/toolStep
If the toolStep is very fine, this can take very long to process...


The algorithm for ball finish is
* Take n=((r+offset)*2+1)/toolStep slices along main progress direction
* Take the highest path of the slice
* offset the slices, according to the position of the slice on the mill path
* intersect the n slices, and keep the highest past
* a rotating buffer contains the n slices, to avoid re-computing at each new slice.



There remain some bugs in rough slice, and in contours.

DodoLaSaumure:
I'll put some corrections to the rough slice and contours soon.
PIerre

DodoLaSaumure:
Hi Tomas, hi all

Here's an alpha release of the stlSlicer plugin.
I know that is not perfect, but given the time I am able to spend on bCNC, I consider it's enough at the moment for my personal use, for now.
Maybe one will have to rework his part sometimes before applying this plugin.
Please give some feedback here.


Here's a render that I was able to make on camotics, with the 3 passes combined (I used a ball end mill for the 3 passes, since it is not possible to combine different tools in camotics.

image

DodoLaSaumure:
> I used a ball end mill for the 3 passes, since it is not possible to combine different tools in camotics.

Ah yes it is, need to do a Tn M6 at each file beginning, where n is the tool number. ;)

bCNC/lib/bpath.py
```diff
@@ -467,22 +467,45 @@ def order(self, P):
...
+		t = Dt/DD #is this sure ? we have sin(AC,CM) / sin(AB,CD) see _intersectLineLineNEW (not used)
+		P = self.AB*t + self.A # is this sure in all cases ?
 		if self.minx<=P[0]<=self.maxx and other.minx<=P[0]<=other.maxx and \
-		   self.miny<=P[1]<=self.maxy and other.miny<=P[1]<=other.maxy:
+		   self.miny<=P[1]<=self.maxy and other.miny<=P[1]<=other.maxy:#is this sure??? what does a<=x<=b return ? Should't we write a<=x and x<=b instead ?
 			return P,None
 		return None,None
 
 	#----------------------------------------------------------------------
+	# Intersect a line with line (not used yet..)
```
Harvie, line 488: You complain about `_intersectLineLine` not being explained enough, but you don't provide a single line of explanation on how ` _intersectLineLineNEW` works nor how it is different from the original method...

bCNC/lib/bpath.py
```diff
@@ -823,18 +846,18 @@ def distPoint(self, dist):
 				return segment.distPoint(dist)
 			dist -= segment.length()
 
-        #----------------------------------------------------------------------
-        # Return linearized path (arcs are subdivided to lines)
-        #----------------------------------------------------------------------
+		#----------------------------------------------------------------------
+		# Return linearized path (arcs are subdivided to lines)
+		#----------------------------------------------------------------------
```
Harvie, line 851: Why use two level indentation comments for methods with single level of indentation? (This is in multiple places)

bCNC/lib/bpath.py
```diff
@@ -1119,6 +1143,52 @@ def invert(self):
 		self.extend(new)
 	reverse = invert
 
+	#----------------------------------------------------------------------
+	# Split path into contours
+	# This not only SPLITs path to contours,
+	# it also takes unsorted segments and JOINs them to closed loops if possible
+	# duplicate of split2contours ? but is does not affect the path, it makes a deepcopy of it
+	# not used yet
```
Harvie, line 1151: if only difference from split2contours is the deepcopy, why not just do deepcopy and then call split2contours? I would really like to prevent duplicate code in bCNC core, because maintenence is already hard enough.

bCNC/lib/bpath.py
```diff
@@ -1805,6 +1878,44 @@ def convert2Lines(self, minlen):
 			if segment.length()<=minlen:
 				segment.change2Line()
 
+	#----------------------------------------------------------------------
+	# Convert Arc to LINES  in a path
```
Harvie, line 1882: There is already `def linearize()` why you need another method for the same thing???

bCNC/lib/bpath.py
DodoLaSaumure: you're right Tomas
_interectLineLineNEW was just written for me, but I wanted to keep it somewhere on my branch I can remove it

bCNC/lib/bpath.py
DodoLaSaumure: Yes I rewrote split2contours for me and wanted to save it somewhere, I can remove it now

bCNC/lib/bpath.py
DodoLaSaumure: OK I did not see linearize function.
=> Can remove it.

DodoLaSaumure:
By the way, I have made further tests and discovered new bugs....

Harvie:
Ok, if you think that some of your methods in bpath are objectively better than the current ones, then you should explain why they are better and it makes sense to use them instead, or fox the original ones. I think the bpath is perhaps the most important part of bCNC and i want to keep it clean, but it is also important to improve and extend it with quality code, so i fully support making sensible changes. And i agree it needs more documentation/explanations in the comments.

DodoLaSaumure:
No I do not think they are better. I used them mainly for debugging and for understanding the code.
Can be removed.
Regards
Pierre

Harvie:
So bpath has to stay more or less sane. But on the other hand if you are author of plugin, i am OK with you puting any crazy experimental code you wish in your plugin as long as it does not completely prevent bCNC from starting. (sometimes this can happen because of syntax errors or python version incompatibility)

DodoLaSaumure:
> So bpath has to stay more or less sane. But on the other hand if you are author of plugin, i am OK with you puting any crazy experimental code you wish in your plugin as long as it does not completely prevent bCNC from starting. (sometimes this can happen because of syntax errors or python version incompatibility)
> 
> bpath is meant like dependable base that everyone can build their crazy plugins on :-) good thing is that most plugins are in bCNC tree, so if we introduce some little incompatibility in bpath, we can easily fix all plugins.

I have restored bpath.py
Main program starts with python2, python3, python3.8
But there remains bugs in the plugin, not crashing the program, but giving strange results ... ( it is experimental, and by chance other developers will help development)

Thanks Tomas
Have a nice day
Pierre

bosd:
What is the status of this one?
@LittlePierre Could you silence some of the codefactor warnings?

LittlePierre:
@Harvie Could you please merge ?

LittlePierre:
it appears to me to be ok for an alpha-test, with these last changes.
Pierre

Harvie:
Thanks!

LittlePierre:
Thank you @Harvie

PR #1565, opened 2021-03-09
Rework shebangs and encodings for Python 3

MicroJoe:
Since [Python 2 is EOL'ed](https://www.python.org/doc/sunset-python-2/), the files can be adapted to properly work with Python 3 by:

1. Using `/usr/bin/env` instead of `/bin/env` to respect common rootfs architecture on GNU/Linux distributions.
2. Using `/usr/bin/env python3` instead of `/usr/bin/env python` to be sure to invoke Python 3 on all distributions.
3. Removing `coding: ascii` since default coding in Python 3 is `utf8` which is compatible with `ascii`.


Harvie:
* Python2 is still supported and will be supported in 0.9.15 release of bCNC. After that it will be dropped and following development will abandon python 2. But it is reccomended to run bCNC using `pythonX -m bCNC` so shebangs are not really needed i guess.
  * I am not sure why there are shebangs in files that are not executables...

MicroJoe:
> Python2 is still supported and will be supported in 0.9.15 release of bCNC. After that it will be dropped and following development will abandon python 2. But it is reccomended to run bCNC using pythonX -m bCNC so shebangs are not really needed i guess.

ACK.

> I am not sure why there are shebangs in files that are not executables...

Agreed. I think this is because the [template in the wiki](https://github.com/vlachoudis/bCNC/wiki/Tutorials:-How-to-create-a-plugin) suggests this format. I will rework this PR to remove the unrequired shebangs instead.

MicroJoe:
Created a new, distinct PR to remove the shebangs. See #1566

PR #1566, opened 2021-03-09
plugins: remove unecessary shebangs

MicroJoe:
The files are not executable, hence the shebangs are not necessary.


Harvie:
Great. Thank you.

PR #1567, opened 2021-03-09
bCNC.desktop: add Keyword key

MicroJoe:
The `Keywords` entry is recommended to be present in `*.desktop` files. We can add a few keywords as a first start.


Harvie:
Thanks

PR #1569, opened 2021-03-09
Add missing __init__.py files

MicroJoe:
These files are necessary for the submodules to be detected by the Python interpreter (`from bCNC.controllers import GRBL0`) — as well as to be automatically detected by the `find_packages()` function in `setup.py` that will now properly detect `bCNC/plugins/` and `bCNC/controllers/` as proper Python submodules.


Harvie:
Are you sure this is a good idea? Plugins and controllers are explicitely scanned and loaded by python code of bCNC, not automaticaly by python.

Harvie:
They contain overlapping definitions and it is important not to import them all at once.

MicroJoe:
After testing locally this works with the plugins aswell.

GitHubCaps:
I may be missing the point here, but plugins are loaded through exec and are not meant to be found otherwise. bCNC checks for them implicitly, hence why they don't have an __init__, controllers could use one (they do get imported), but plugins should be kept the way they are and not checked through setup (imho).

GitHubCaps:
For you to better understand why they don't have an init, bCNC directly put them in path.

```python
sys.path.append(os.path.join(PRGPATH, 'lib'))
sys.path.append(os.path.join(PRGPATH, 'plugins'))
sys.path.append(os.path.join(PRGPATH, 'controllers'))

how plugins are loaded:
```python
		# Find plugins in the plugins directory and load them
		for f in glob.glob("%s/plugins/*.py"%(Utils.prgpath)):
			name,ext = os.path.splitext(os.path.basename(f))
			try:
				exec("import %s"%(name))
				tool = eval("%s.Tool(self)"%(name))
				self.addTool(tool)
			except (ImportError, AttributeError):
				typ, val, tb = sys.exc_info()
				traceback.print_exception(typ, val, tb)
```

As, far as the lib dir, it is meant for bCNC and 3rd party libs (the init file came from python_utils or stl), but should be considered private to bCNC.

One of the things I have done to improve the project (not posted yet) is to separate the 3rd parties to make it a little easier to know which code is native.

Harvie:
So do you think it does make sense to close this without merging?
are there some other benefits we can achieve by having these init files?

MicroJoe:
I think it makes sense to close without merging, thanks for the explanation.

PR #1570, opened 2021-03-09
introduce bCNC.1 man page

MicroJoe:
A man mage is always nice to have to understand what a program is and how it can be invoked. It is a start that can then be improved to add more details if needed.

### DIFF
--- a/bCNC.1
+++ b/bCNC.1
@@ -1,0 +1,65 @@
+.TH BCNC "1" "March 2021" "bCNC" "User Commands"
+.SH NAME
+bCNC \- GRBL CNC command sender, autoleveler and g-code editor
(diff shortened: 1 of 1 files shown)


Harvie:
Interresting idea... Do you think it would be possible to generate manpage directly from github wiki page?

MicroJoe:
> Interresting idea... Do you think it would be possible to generate manpage directly from github wiki page?

From the wiki I do not know, as it would need to have internet access and so on… but from a markdown file or something else it should be possible. See [rst2man](https://manpages.debian.org/testing/docutils-common/rst2man.1.en.html) for example. Maybe the wiki content should be moved to files into the repository so that documentation can be used everytime — even offline.

PR #1573, opened 2021-03-16
Modify driller plugin to work with Eagle's Exellon

kswiorek:
The plugin behaved very strangely when importing .xln files generated from Autodesk Eagle.
The number of decimal places is now read from the beggining of the file instead of the number of digits, which isn't constant with Eagle.
Also added a tool change command if the machine isn't a laser cutter.


Harvie:
Looks reasonable. Thanks!

PR #1580, opened 2021-04-06
remove unused shebangs

MicroJoe:
Some shebangs are used in files which are not executable. Thus they are not useful and should be removed.


Harvie:
Thanks

PR #1584, opened 2021-05-06
Support up to 8 axis from the device status report (X,Y,Z,A...E)

gcobos:
This allows to control an arbitrary number of axis if needed in a clean way. 

I have been testing my port of GRBL to esp8266 using bCNC for a while already. My port handles 8 axis instead of 3, otherwise is compatible with GRBL

### DIFF
--- a/bCNC/CNC.py
+++ b/bCNC/CNC.py
@@ -46,6 +46,8 @@ YZ   = 2
 CW   = 2
 CCW  = 3
 
+AXIS_NAMES = ["X", "Y", "Z", "A", "B", "C", "D", "E"]
+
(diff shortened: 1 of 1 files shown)


Harvie:
GRBL currently support 6 axes. Is anyone actualy using more than that?
Also correct me if i am wrong, but after inspecting the code, it seems that you've modified only small part of the CAM module (namely code needed to move g-code block).
Is there point of having 8 axis support in this very specific use case?

gcobos:
I'll close this for now. It might be that it was useful for me, but not needed anymore. If that's not the case, I'll make the changes required and open a new request.

PR #1610, opened 2021-08-19
DRO frame feed rate display

calabr:
Current feed rate set field locate in State frame and only accessible from Control window.
Jogging move do not display actual feed rate

Feed rate display added to DRO frame
Allow to see actual feed rate during any machine move, including jogging.
DRO frame buttons slightly re-arranged to feet additional field.
Fed screen


Harvie:
This seems as a nice idea. Can you please update your branch so it can be merged?

Issue #1659, opened 2022-01-22
PIP version of bCNC does not run with python 3.10.2 on windows 10


cshunt:
After installing bCNC with python -m pip install bCNC, trying to run bCNC.bat or bCNC.py fails. "py -m bCNC" from cmd prompt.

\AppData\Local\Programs\Python\Python310\lib\site-packages\bCNC\Utils.py
NameError: name 'sys' is not defined

It appears this issue has been resolved in the master branch but not in a release yet. "Import sys" is missing from several files. It looks to be related to an update in python. utils.py, toolspage.py & filepage.py
adding "import sys" in these files lets bCNC run.. not sure if there are other modules that would crash because of this

Git-Bruno:
I ran 'pip install bCNC' a whille back without problem with python 3.8 but I also get the error NameError: name 'sys' is not defined when I start bCNC with python 3.10 after installing bCNC with 'pip install bCNC'.
With 3.8 there is a directory  bCNC-0.9.14.317-py3.8.egg-info in Python38\Lib\site-packages\ after the install and there is an 'import sys' in modules utils.py, toolspage.py & filepage.py.  

With 3.10 there are errors compiling numpy (cl.exe failed with exit status 2 & ERROR: Failed building wheel for numpy) and when the pip command exists there is a bCNC-0.9.14.307-py3.10.egg-info in Python310\Lib\site-packages\. 
While the pip command runs I see what I interpret as an attempt to install different releases of bCNC down to bCNC-0.9.14.307 in which the 'import sys' statements are missing in modules utils.py, toolspage.py & filepage.py.  

After adding 'import sys' in modules utils.py, toolspage.py & filepage.py helps but then this is to run an old release of bCNC. No update done.
Below the last error messages I get. 
Collecting bCNC
  Using cached bCNC-0.9.14.316.tar.gz (1.3 MB)
  Preparing metadata (setup.py) ... done
  Using cached bCNC-0.9.14.315.tar.gz (1.3 MB)
  Preparing metadata (setup.py) ... done
  Using cached bCNC-0.9.14.314.tar.gz (1.3 MB)
  Preparing metadata (setup.py) ... done
  Using cached bCNC-0.9.14.313.tar.gz (1.3 MB)
  Preparing metadata (setup.py) ... done
  Using cached bCNC-0.9.14.312.tar.gz (1.3 MB)
  Preparing metadata (setup.py) ... done
[...]

PR #1660, opened 2022-01-22
Update setup.py

cshunt:
update opencv_version, version listed is not available and causes failure to install.


cshunt:
See issue #1659 for context

Harvie:
Thank you

Harvie:
Pushed to pypi as bCNC-0.9.14.318

Issue #1663, opened 2022-01-25
Mesh pluging not loaded with python 3.10


Git-Bruno:
The slicemesh plugin fails to load when bCNC runs with 3.10. From the messages received at bCNC startup this error is caused by a change in python 3.10 : 

`
Traceback (most recent call last):
  File "C:\Python310\lib\site-packages\bCNC\ToolsPage.py", line 1099, in __init__
    exec("import %s"%(name))
  File "<string>", line 1, in <module>
  File "C:\Python310\lib\site-packages\bCNC\plugins\slicemesh.py", line 38, in <module>
    import stl #FIXME: write smaller STL parser
  File "C:\Python310\lib\site-packages\bCNC\lib\stl\__init__.py", line 1, in <module>
    from .stl import BUFFER_SIZE
  File "C:\Python310\lib\site-packages\bCNC\lib\stl\stl.py", line 12, in <module>
    from . import base
  File "C:\Python310\lib\site-packages\bCNC\lib\stl\base.py", line 77, in <module>
    class BaseMesh(logger.Logged, collections.Mapping):
AttributeError: module 'collections' has no attribute 'Mapping'
`

The slice mesh plugin can be loaded if I change 'import collections' to 'import collections.abc as collections' in base.py, and stl files can be loaded too. 
This change works also for python 3.8 but maybe there are more side effects down the road requiring more changes than just changing the import.

imechura:
I cannot tell what is implied but posting this link. Is it fixed? I just did a clean install using pip install bCNC and I get this error. Is the link implying that I should merge the changes into my source code or is there another release not supported by PIP that need to be installed?

PR #1673, opened 2022-02-08
Import ABC from collections.abc for Python 3.10 compatibility.

tirkarthi:
Fixes https://github.com/vlachoudis/bCNC/issues/1663


GitHubCaps:
Good work, but should be placed in place of collections (with standard imports, not below the locals), just my 2 cents

bosd (approved):
Code review! LGTM :+1:

Harvie:
Thanks for the fix and the review...

PR #1677, opened 2022-02-20
Implement rendering dotted fonts as a path for drilling

fedulvtubudul:
I've added an option for TEXT plugin that allows to use dotted fonts for drilling (not milling) the text. That can be useful for production of templates, nameplates, front panels, etc.

I'm using free dotted fonts from here: https://www.urbanfonts.com/fonts/dotted-fonts_page-1.htm.

Here is an example of generated path:

Screen Shot 2022-02-20 at 20 37 31


Harvie:
Seems to be a nice idea, can you please resolve the conflicts for merging? I beleive there were some changes in text plugin recently. Added support for new font format if i remember correctly.

Issue #804, opened 2018-04-15
Zooming Viewport with the mousewheel only works in one direction


zirkeltraining:
Thank you for your software, it's great to get into CNC. 

A minor annoyance: 
on a Mac, OSX 10.11.6, using Python2, (the only config that i tested),
i cannot zoom OUT with the scroll wheel of my mouse (or two-finger-swipe-down). It only zooms into one direction, IN.  That's kind of frustrating because it's what every CAD tool uses, and navigating the viewport is clumsy without it.

vlachoudis:
I don't have a mac to test, but its strange since up to now I've got the impression that it was working on mac.

Harvie:
Works for me. Do you still have this problem with lastest master?

fedulvtubudul:
Issue is still here and it's really boring. I've tried to debug and it seems related to Python 2/3 rounding rules differences. Please check this PR: https://github.com/vlachoudis/bCNC/pull/1678. Works for me, but I've only tested on Mac.

PR #1678, opened 2022-02-20
Fix mouse wheel zooming in both directions on Mac

fedulvtubudul:
This PR fixes https://github.com/vlachoudis/bCNC/issues/804

Here is original zoom wheel handler code:
```python
def wheel(self, event):
    self.zoomCanvas(event.x, event.y, pow(ZOOM,(event.delta//120)))
```

`event.delta` equals -1 (scroll up) or 1 (scroll down) on my setup (MacOS 12.2.1, Logitech MX 3 mouse, medium scrolling speed in system settings, "natural" scroll direction). In Python 2 rounding goes away from zero, so `event.delta//120` gives either 1 or -1, which is OK. But running in Python 3.9, `event.delta//120` results -1 or 0. That why only one zooming direction works.

I propose a fix that seems compatible both with Python 2 and 3 and should not break or change scrolling behavior.

### DIFF
--- a/bCNC/CNCCanvas.py
+++ b/bCNC/CNCCanvas.py
@@ -1008,7 +1008,9 @@ class CNCCanvas(Canvas, object):
 
 	# ----------------------------------------------------------------------
 	def wheel(self, event):
-		self.zoomCanvas(event.x, event.y, pow(ZOOM,(event.delta//120)))
+		scaledDelta = event.delta / 120
+		deltaPower = math.ceil(scaledDelta) if scaledDelta > 0 else math.floor(scaledDelta)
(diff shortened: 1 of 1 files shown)


Harvie:
Thank you for this fix. I hope few other users can test this and share their experience with fix as well.
If i remember correctly, i've merged similar thing some time ago and it ended up breaking zoom for some of the users (while fixing it for others). So i will test it myself on ArchLinux/X11 and hopefully we can get feedback from users on other platforms as well.

rschell (approved):
Wasn't having issues on my Pi implementation, zoom function continue to work both locally and in a remote VNC session.

Harvie:
Ok thanks, can you please resolve the conflicts, so we can merge?

PR #1701, opened 2022-05-08
Save ProbeFrame.autogoto setting

pci06:
Hello,
This setting is always enabled and in some circumstances could lead to big surprise when there's no previous probe location.
This way the user can decide.

Regards,
Pierre.


Harvie:
Can you please explain? does this implement saving state of autogoto checkbox between bCNC restarts?

Harvie:
Seems reasonable, can you please rebase to latest version, so i can merge? Thanks in advance.

pci06:
Hello,
New PR #1877 as file diverge too much.
Regards.

Issue #1509, opened 2021-01-08
Web Pendant, Picture flickering


Guenni75:
Happy New year.
I have a Problem with the Web Pendant.
It is the same on Windows, Linux, Android and every Browser i can use.
bCNC is installed on Lubuntu.

https://user-images.githubusercontent.com/14917841/104018878-dbe64400-51ba-11eb-83a2-b44c898ca375.mp4

GitHubCaps:
@Guenni75 Have you tried directly from bCNC first?

Make sure that you are connected to pendant and can run some code locally via bCNC through the pendant.

Personally, I stopped using Lubuntu a long time ago because of issues with camera, web, etc. I found Linux Mint XFCE versions to fit the bill a bit better, I use the 32-bit versions for my old machines and of course the 64-bit for the newer. I stay away from Cinnamon and Mate, because I prefer lightweight OSes myself.

GitHubCaps:
@Guenni75 Did some further investigating on my end and it does have an issue when opening a file in pendant. It works in Python2, however it needs patching for Python3.

I will reply back as soon as I have a patch for it.

GitHubCaps:
@Guenni75 and anyone else that would like to test the following before a commit.

Tested in python 2.7.14 (already worked there) and 3.8.2 (lots of issues there)
[Pendant.py.zip](https://github.com/vlachoudis/bCNC/files/5792214/Pendant.py.zip)

`Pendant.py` for those that don't like to download anything suspicious:
```python
# -*- coding: ascii -*-
# $Id$
#
# Author: Vasilis Vlachoudis
# Email:  vvlachoudis@gmail.com
# Date:   24-Aug-2014

from __future__ import absolute_import
from __future__ import print_function
__author__ = "Vasilis Vlachoudis"
__email__  = "Vasilis.Vlachoudis@cern.ch"

import os
import re
import sys
#import cgi
import json
#import urllib
import tempfile
import threading

# ---------------------------------------
# For debugging issues
# ---------------------------------------
import traceback
from errno import EPIPE
try:
	broken_pipe_error = BrokenPipeError  # Python3 has: BrokenPipeError
except NameError:
	broken_pipe_error = IOError
# ---------------------------------------

try:
	import urllib.parse as urlparse
	from urllib.parse import unquote
except ImportError:
	import urlparse
	from urllib import unquote as unquote

try:
	import http.server as HTTPServer
except ImportError:
	import BaseHTTPServer as HTTPServer

try:
[...]

PR #1705, opened 2022-05-23
Web pendant fixes

LepeshkinSN:
**Brief**
Fixed some problems with web pendant:
1. Wrong (only on Linux?) HTTP content-length value while serving disk-sourced content (canvas and icons)
2. Problems with file upload (seems to be related to Python 3)
3. Complex canvas structure on slow machines cause refresh failure
4. Infinite growth of canvas image cgi argument string

**Details**
This changes addresses  issue #1509 (and maybe some others).
I'am using bCNC 0.9.14 on Debian 11 x64 with Python 3.9.2. My browser is [Version 1.31.91 Chromium: 95.0.4638.69 (Official Build) (64-bit)]. 
Web pendant was not usable for me because there were no canvas image (but sometimes it flickers) and I wasn't able to upload file through it. Here is related problems and fixes:
1. It seems os.path.getsize() on my system returns disk size of file, not content length of file. So, browser gets wrong Content-Length size on small files (smaller than disk block size) and discards it content after connection closing. I've replaced getsize() with seek()/tell() combinations.
2. File upload code seems to be not adapted to Python3. There are some changes in classes and between bytes/strings arguments and variables. 
3. When your cnc code is very long, it tooks a while to generate canvas image. On timeout (1 sec.), old code causes request abortion and generation of the new request, which invoke new canvas generation and this sequence will never stop, and you will see no canvas at all. My changes adds check whether previous canvas image was loaded before starting new request.
4. Canvas image update code adds current time as argument to canvas image url. Maybe it's bug in browser, but in such case URL of canvas image grows infinitely. I've hard-coded pre-argument part of URL and it seem's to work.


rar8000:
I merged these changes manually to a local clone of the current release and to my PR. It works great. For large images, the pendant does not show images without these changes. 

Thanks!

PS: Since this work was done before the python 3.8 port. I had to do the merge manually in an editor.

Harvie:
>  It works great

Thanks for the review

Unfortunately i cannot merge this PR, since there's conflict

LepeshkinSN:
Hi!
Could you explain what conflict it is?
There were problem with codefactor check, but now it's resolved.
But it seems there are some other conflicts. On "Resolve conflicts" page there are some chunks of code, but they differ only in indention size. Payload is the same. What's wrong with it?

Harvie:
It's not really about codefactor. it's simply the fact that some of the files you are trying to modify had been also changed in vlachoudis/bCNC since you've forked and therefore the changes in your fork cannot be applied, you need to merge vlachoudis/bCNC first and manualy resolve the conflicts in order to not overwrite changes done in upstream by other PRs...

LepeshkinSN:
Hmm...
If so, why it was not merged a year ago, when I've created this pull request?
My repository was cloned and modified the same day. Furthermore, next changes to Pendant.py are introduced only at 1 Sep. 2022, while my pull request is from 23 May 2022.
BTW, at that time, there were codefactor check fail, but it refers to original chunk of code. There were no check fails in my changes.
I guess story is this: a year ago I've made changes to original code. At this moment original code were not conform to codefactor, but codefactor checks for pull requests were already on. Because of that my pull request were not merged automatically and left for manual merge by repository owners. Unfortunately, they didn't manage to merge it in time. May be they even didn't look on it. Instead of this, they introduced some code cleanup and prettier. And now there is conflict between old code and new "pretty" one.
Well, ok. Now I have no time to deal with this new pretty code and you have to do it by yourself.
Cheers!

Harvie:
Maybe there were more PRs affecting the web pendant. i merged the first one and this one become conflicted immediately, therefore i've waited for a year if you notice. I understand your frustration, but that's just how things are when multiple people create PRs affecting same files. However it seems this conflict might not be really hard to resolve, since the changes are reasonably small.

Maybe @rar8000 can give us some insight how he got this merged when he's been testing it.

rar8000:
Hi. Basically I went line by line and copy pasted. The changes don't conflict to the human eye, but the line numbers have been shifted; so git cannot merge. I cloned a fresh updated bCNC, then manually found each line via github interface. Then copy and pasted. Took less than 5 mins. I attached the two updated files in a zip.

[pendant.zip](https://github.com/vlachoudis/bCNC/files/12185694/pendant.zip)

rar8000:
I also merged the changes to my "test" branch. Different than my release branch. This fix gets the pendant working for me. 

Other errors are still thrown when the pendant is running, but they seem to be caught and displayed. These errors are separate from this fix. With the fix enabled, the pendant works, though a bit slow and can periodically hang the UI updates to the main program while running.

bCNC/Pendant.py
```diff
@@ -112,9 +112,11 @@ def do_GET(self):
 		elif page == "/icon":
 			if arg is None: return
 			filename = os.path.join(iconpath, arg["name"]+".gif")
-			self.do_HEAD(200, content="image/gif", cl=os.path.getsize(filename))
 			try:
 				f = open(filename,"rb")
+				f.seek(0,2)
```
Harvie, line 117: why are you calling the seek() ?

LepeshkinSN: Hi!
Sorry, but I don't remember precisely why I placed this seek() here. I think without this seek() it starts to read file somewhere at different location, but not at beginning of the file. One thing I can surely say - it was on Debian Linux 10 (buster) and python 3. I think you can try to remove this line and check how things will work.

Harvie: According to https://www.geeksforgeeks.org/python-seek-function/
`f.seek(0,2)` does jump to the end of file. this makes me slightly suspicious about this PR :-)

LepeshkinSN: Oh, now I remember. There were problem with os.path.getsize() - IDK why, but it returning wrong value.
So this is workaround - seek to the end of file and get current position to determine file length.

Harvie: Ok, makes sense... Can you please put this explaination in form of code comments?

Harvie: It's occuring in several places. perhaps preferred way imho would be to create separate file length function that would:

1. store current position
2. seek to end and store file length to variable
3. seek back to initial position
4. return file length
5. contain comments about why we do this

LepeshkinSN: I thought you already included this changes (using something different method than merge pull) after 27 Jul 2023.
If no - well, may be this week I will have some time to do it.

Harvie:
There still seems to be some conflict preventing me from merging this...

bCNC/Pendant.py
```diff
@@ -38,6 +38,17 @@
 class Pendant(httpserver.BaseHTTPRequestHandler):
     camera = None
 
+    # ----------------------------------------------------------------------
+    def get_file_size(fh):
```
Harvie, line 42: thanks

Harvie: Maybe it's related to this? https://bugs.python.org/issue41092

LepeshkinSN:
> There still seems to be some conflict preventing me from merging this...

I can't help you in this situation - GitHub shows me "Able to merge. These branches can be automatically merged.", "All checks have passed", "This branch has no conflicts with the base branch". All green marks.

bCNC/Pendant.py
Harvie: Perhaps `os.stat(path).st_size` would give better results?

Harvie:
> I can't help you in this situation - GitHub shows me "Able to merge. 

oh sorry, when i switch from "rebase" to "merge". it's suddenly green.

bCNC/Pendant.py
LepeshkinSN: May be... But I'am not a professional Python programmer, so I don't know exact details about that.

Harvie:
Weird,
i've tried this:

```python
import os

print(os.path.getsize("test.jpg"))
print(os.stat("test.jpg").st_size)
```

and it does return same number as `wc -c test.jpg`. perhaps it's some issue specific to your version of pyhton...

Harvie:
Maybe you can also try `os.fstat(fd).st_size` which works directly on the open file descriptor rather than path...

LepeshkinSN:
Yes, I've checked os.path.getsize() - now it returns correct value. Maybe
it's better to use this function instead of my workaround.
But at the time I created this pull request it was way I could do it
function properly.

ср, 14 февр. 2024 г., 14:45 Tomas Mudrunka ***@***.***>:

Harvie:
> Yes, I've checked os.path.getsize() - now it returns correct value. 

now i am quite confused. i thought that you were doing this workaround, because os.path.getsize() is not working for you... what had changed in the meantime? have you upgraded your python version or something?

LepeshkinSN:
I think minor version number of Python 3 on my system was changed since
then.
Periodically I perform system update from standard Debian repo, and it
upgrades pretty much packages. So it is very likely Python 3 was among
them.

ср, 14 февр. 2024 г., 20:15 Tomas Mudrunka ***@***.***>:

PR #1740, opened 2022-09-01
Use Subprocess instead of os

bosd:
An quick attempt to silence B605: start_process_with_a_shell.

```
This rule looks for the spawning of a subprocess using a command shell. This type of subprocess invocation is dangerous as it is vulnerable to various shell injection attacks. Great care should be taken to sanitize all input in order to mitigate this risk. Calls of this type are identified by the use of certain commands which are known to use shells.
```

### DIFF
--- a/bCNC/bmain.py
+++ b/bCNC/bmain.py
@@ -5,6 +5,7 @@
 # Date: 24-Aug-2014
 
 import os
+import subprocess
 import socket
 import sys
 import time
@@ -2572,7 +2573,7 @@ class Application(Tk, Sender):
         CNC.vars["_OvChanged"] = True  # force a feed change if any
         if self._onStart:
             try:
-                os.system(self._onStart)
+                subprocess.call(self._onStart, shell=False)
             except Exception:
                 pass
 


bosd:
Just tested this one on python3.10 on windows 10.
Working ok.

rschell (approved):
Tested on 3.11 on linux (rapsberry pi 4 with linux 5.15.65 64-bit)

Harvie:
Can you please explain how is this "safer" ?

bosd:
Sorry, I cannot.
I'm not very knowledgable on the subject.
Basically this pr is to silence a "critical" codefactor.io complaint.
I'm utiliting the suggested fix from codefactor.io.

I'll paste here the complaint and explanation.

This rule looks for the spawning of a subprocess using a command shell. This type of subprocess invocation is dangerous as it is vulnerable to various shell injection attacks. Great care should be taken to sanitize all input in order to mitigate this risk. Calls of this type are identified by the use of certain commands which are known to use shells.

Example of insecure code:
```
import os

os.system('/bin/echo suspicious code')
```

The **subprocess** module provides more powerful facilities for spawning new processes and retrieving their results; using that module is preferable to using this function.

Example of secure code:

```
import subprocess

subprocess.call('/bin/echo suspicious code', shell=False)

```

Further Reading

[The Python Standard Library - subprocess - Frequently Used Arguments](https://docs.python.org/2/library/subprocess.html#frequently-used-arguments)


[Bandit - B605: start_process_with_a_shell](https://bandit.readthedocs.io/en/1.7.4/plugins/b605_start_process_with_a_shell.html)

bosd:
@Harvie Can you merge this?

bosd:
Ping @Harvie

bosd:
Ping @Harvie

Harvie:
Yeah, i am still not sure whether this is actualy making something safer or just removing actual useful feature, because some automated tool thinks it might be unsafe without understanding context...

bosd:
Did'nt notice any feature removal.. if so, we an always revert.

rschell (changes requested):
Need to include the self._onStop call made in sender.py, as well, if incorporated.

Harvie:
i think this will break functionality unless shell=true is provided. When shell=false is provided, it expects array of arguments rather than string.

PR #1741, opened 2022-09-02
New control frames revised

rschell:
I took my previous New_Control_Frames pull request and squashed/merged the commits and applied them the the new Cleancodev2 version of bCNC.

Here are two new control frames, both using slider controls to adjust the step size or step rate. The first frame adds fixed step size sliders for the Z and XY axes. The second frame adds a continuous control frame using a single slider to adjust the step rate for all three axes. Either or both can replace the current control frame by making the appropriate selection in the bCNC startup files.

Ron


m1ch:
@rschell Looks really nice! I like especially the continuous one! 

Might I suggest, that you add three radio-buttons at the top of the control-frame, that you can select the mode directly.

rschell:
> @rschell Looks really nice! I like especially the continuous one!
> 
> Might I suggest, that you add three radio-buttons at the top of the control-frame, that you can select the mode directly.

While I have created labeled buttons to toggle the frames, using pack() and pack_forget() does not collapse the space taken by the frame.  

Tried destroying the frame, but there is no clean way of doing that since the code for the frames is not consolidated within the control page code.  For example, bmain.py contains the key bindings for the frames.  This is complicated by the fact not all frames are created bmain.py during initialization. 

Not sure I see a way to implement this yet.

Harvie:
Any screenshots pls?

rschell:
> Any before/after screenshots pls?

slider control:
slider_control_frame

Continuous Control:
continuous_control_frame

Harvie:
Will the keyboard control by +-*/ keys still work?

Harvie:
So you have two modes and ability to switch between them? what is the continuous control?

rschell:
The original bindings continue to work even though the control has not been displayed in my case, my usage has been more mouse based.  So without the control displayed, the jogging distance using the keyboard is hidden.

 

Ron

rschell:
> So you have two modes and ability to switch between them? what is the continuous control? My machine really can't have Z and XY joging set to the same speed, because Z is much shorter and that's gonna lead to crash immediately.
> 
> Also i usualy use touchpad in the workshop, because i don't want to have mouse in the sawdust. and setting sliders with touchpad is quite fiddly. Why not keep the buttons next to the sliders? there's still enough room.

I will see if I can insert the sliders into the original control.

rschell:
would something like this be better?
Modified_Original
still need to connect the slider to the displayed values and vs versa

rschell:
> still need to connect the slider to the displayed values and vs versa

I have added sliders as shown above and linked them to the combo boxes and results of the keyboard bindings.  The resulting frame makes mouse use easier yet retains the keyboard functionality.

Should I submit the modifications to the original control frame separately from the continuous control or keep them combined?  I can squash the PR or submit a new one if that would be appropriate.

Since the proposed slider control frame features are now incorporated in the original control frame, that portion of the commit would be dropped.

rschell:
Will the keyboard control by +-*/ keys still work?I did not change the bindings of the original control, if that control is not selected/displayed, I’m not sure what will happen.  Will check, been thinking that some of the binding code needs to be relocated to the the appropriate frame code.  Haven’t thought though all those implications.I have both the slider and continuous controls displayed and alternate between them depending on which type of jog I desire.  The continuous control does not have a specified distance, the machine will jog as long as mouse holds the button down, movement stops when the button is released.  Movement will stop at wpos=0.  If you like, the original control and the continuous control can be displayed and used similarly.RonOn Sep 18, 2022, at 2:21 AM, Tomas Mudrunka ***@***.***> wrote:﻿
So you have two modes and ability to switch between them? what is the continuous control?

PR #1748, opened 2022-09-17
fix push reloacted parentheses

rschell:
Not sure how this happened was fine locally


Harvie:
thanks :-)

bosd:
What happend? Build on this one shows a red cross

Harvie:
Not sure, but the red cross from codefactor is caused by lines not modified in this PR... so perhaps some cache expired at codefactor side?

PR #1749, opened 2022-09-22
Add jograte and sliders to Control Frame

rschell:
Attempt number 3 (squashed, cleaned, triple checked), removed continuous frame from PR (can be submitted in another PR, if requested):

Allow user to specify the jog rate rather than default to the maximum machine jog rate.  Values above the maximum machine jog rate will be capped by the controller.  Only positive float numbers are allowed, field must always validate to a float number (blank not currently allowed).

Added Z and XY slider controls to permit easier distance adjustments via a mouse.  The sliders have user defined stops along a logarithmic scale.  Adjustments by the sliders are reflected in the combobox values.  Adjustments via the keyboard +-/* strokes are mirrored in the
sliders.  The Z distance controls can be linked via the Z combobox XY selection.

Also improved exception handling throughout.
 
Original with Jograte

Changes to be committed:
        modified:   bCNC/ControlPage.py
        modified:   bCNC/bCNC.ini
        modified:   bCNC/controllers/GRBL1.py


rschell:
Need to add a default feed rate to the jog function definition, so calls to jog in the abc frame don't need to be modified.  Should add a similar jog rate control to the ABC frame.

I have a patch on my local git version but not sure what the result of a push will be since other commits have been added since the original push and I rebased my local copy.

PR #1755, opened 2022-11-01
Fixed "Package would be ignored" warning

yarda:
This is due to https://peps.python.org/pep-0420/

Reported in:
https://bugzilla.redhat.com/show_bug.cgi?id=1684603#c17

Signed-off-by: Jaroslav Škarvada <jskarvad@redhat.com>


Harvie:
thanks

PR #1771, opened 2022-12-14
Update CNC.py

MARIOBASZ:
Solve problem pocket operation in narrow spaces


MARIOBASZ:
The proposed solution (just modify one line) is not to the liking of Code Factor, I imagine it will not be adopted.
If someone wants to provide a better solution, it will be useful for many.

PR #1790, opened 2023-01-02
Fixes to Profile when pocket is enabled

MARIOBASZ:
1- Correct profile with pocket activated in multiple selection.
2- Correct stepover parameter for pocket .
3- Add updown field in pocket to allow to be placed in up position
4- The pocket operation is now before the profile operation.
5- On the canvas, the generated profile and pocket operations are selected

### DIFF
--- a/bCNC/CNC.py
+++ b/bCNC/CNC.py
@@ -4409,8 +4409,13 @@ class GCode:
         # and profile
         if pocket:
             msg = msg + self.pocket(
-                newblocks, abs(offset), CNC.vars["stepover"] / 50, name, True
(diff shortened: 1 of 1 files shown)


Harvie:
Thanks for fixes. Is this thorouhgly tested?

MARIOBASZ:
Only minor fixes were needed.
Works correctly.
Attached test file. It becomes very evident if you select only the circles
[PROFILE WITH POCKET ERROR.txt](https://github.com/vlachoudis/bCNC/files/10332864/PROFILE.WITH.POCKET.ERROR.txt)

Harvie:
Ok, thanks

PR #1800, opened 2023-01-13
Jigsaw plugin: Add deviations in the crossings of the cuts, so that every piece can change in size

gcobos:
It also reduces the radius of the basic taps, to avoid collisions on very small puzzles


gcobos:
@rschell This new PR implements deviations in the crossings of the cuts, effectively changing the sizes of the pieces.I think it's ready for a merge, @Harvie

Harvie:
Thanks

PR #1803, opened 2023-01-19
Update TLO after Stop button is hit

AdamKrovina:
Hitting the Stop button resets the TLO value field shows zero, however the actual TLO value sets correctly set to it's previous value. Calling self.viewParameters() updates the TLO field correctly.


Harvie:
Thanks

MARIOBASZ:
This looks interesting. surely you will use it.

PR #1815, opened 2023-03-05
Give a bit of flexibility to user defined buttons layout

jlirochon:
- Add Buttons.nrows to configure the number of displayed rows
- Fix n+1 problem (store 12 buttons in config instead of 13)
- Factorize LabelGroup.grid*rows()


Harvie:
Can you please provide some before/after screenshots, so we understand what this does? Thank you.

Issue #1661, opened 2022-01-23
How to Call any other executable using bCNC custom Buttons


engrjawadulhassan:
Hello,

Can i call any other executable from bCNC ? using custom buttons , e.g calculator.exe etc.

I was curious about this , as i read the previous issues but found nothing about this.

Thanks

CmdrCody51:
I hope so too. My 3D printer software can (OctoPrint) and I use simple scripts to power lights and printers.
It would be nice to be able in bCNC - turn on lights, mill, vacuum and fans that aren't really specific to GCode.

CmdrCody51:
Looks like I can do what I want through the Pendent web ui. At least now I have time to see if I can find it in the program.

Harvie:
You should be able to embed python scripts in the buttons g-code. See wiki for details.
https://github.com/vlachoudis/bCNC/wiki/Scripting

CmdrCody51:
I sure am glad you verified this before closing it!!!!!
Too bad the example on your wiki doesn't work huh?
I put this in button 9:
%import os; os.system("xterm")
Clicked Ok and it sat there fat dumb and happy.
Guess that makes it more of a BUG.

Harvie:
> I sure am glad you verified this before closing it!!!!!

I am sure glad that you've prepared PR with fix instead of pointing fingers huh :-)
It's not like we are paid or otherwise obliged to check your use case. Just trying to provide free software in our personal free time. bCNC has lots of features and when something gets broken it might not be immediately obvious.

Can **please** you take a look why this is not working? I remember it was working ok in some older version.

CmdrCody51:
Sorry - ya - I got miffed when every thing says it oughta work and it didn't...
First thing I noticed is my version is 0.9.14-dev on RPi with 3.9.2 python.
It was some fix for running on the RPi. Not sure why but I don't even have some of the tabs.
Looks like I have to drill down into my local copy to find what it ought to do as well as why it won't do it.

CmdrCody51:
I see where it oughta run the command.(CNC.py) I guess the editor is croaking on entering the command because it never gets that far.

CmdrCody51:
It is choking in configparser.py in the /usr/lib/python3.9/

CmdrCody51:
Ah man....................
On the first entry in the command window you have to escape the '%' - so you use '**%%**'
If you edit it later you have to use '**%%**'
It is even stored as '**%%**'
You can close this now.....

[...]

Issue #1304, opened 2019-11-13
Button macro: all python code execute before gcode!


BernardG:
I am trying to probe a part to find its middle position.
It looks like both my python code and my Gcode are fine. The only problem is that %wait seems to do nothing, so that, obviously, python variables do not get the proper value.

I understand that python execute locally, while Gcode is sent to GRBL. So I guess, my question is two fold: (1) Would there be a way to make it all python code? (2) Or, is there something to do to make %wait works, OR, any way to force python to wait an OK from GRBL before executing? (I guess that's what %wait is made for anyway...)

Here is my code so far:
```
%_long = app.entry("Entrez la longueur de la pièce: ");
%_larg = app.entry("Entrez la largeur de la pièce: ");
%_flong = float(_long);
%_flarg = float(_larg);
%_posXM = float();
%_posXP = float();
%_posXmi = float();
%_posYM = float();
%_posYP = float();
%_posYmi = float();
%_distmvt = float();
%_diffPos = float();
%msg Debut de Sondage
(==================X=================)
(on assume que Z0 à été fait auparavant)
G91 G0 Z3
%_distmvt = float(0.5 * _flong + 10.00);
%print("distance mouvement = " + str(_distmvt));
G91 G0 X-[str(_distmvt)] (On recule de la longueur /2 + 10 mm)
G91 G0 Z-8 (On descend rapidement de 6 mm)
G38.2 X[str(_flong)] F100 (On avance en X de la longueur (au maximum))
G38.4 X-1 F20 (Après avoir touché, on s'écarte de 1 mm lentement)
%wait;
G4 P5
G10 L20 P2 X0
%_posXM = float(wx);
%print("PosXM = " + str(_posXM));
G91 G0 Z8 (remonte Z là où il était, puisqu'on est en coordonnées relatives)
%_distmvt = float(_flong + 10);
%print("distance mouvement 2 = " + str(_distmvt));
G91 G0 X[str(_distmvt)] (avance de la longueur indiquée + 10 mm)
G91 G0 Z-8 (On descend rapidement de 6 mm)
G38.2 X-[str(_flong)] F100 (On recule en X de la longueur (au maximum))
G38.4 X1 F20 (Après avoir touché, on s'écarte de 1 mm lentement)
%wait;
G4 P5
G10 L20 P2 X0
%_posXP = float(wx);
%print("PosXP = " + str(_posXP));
%print("Longueur trouvée = " + str(abs(_posXM) + abs(_posXP)));
[...]

Issue #1289, opened 2019-10-05
Scripting: How to wait last command to finish to get final XYZ position?


rasciodc:
Not really an issue, but a question:

Using the scripting on the user macro buttons I'm running this code:
`G91 G38.2 X-5 F100
%xl=float(wx)
G91 G38.2 X5 F100
%xr=float(wx)
%center_x=(xl+xr)/2
G91 G0 X(center_x)
G92 X[0]
`

Harvie:
have you tried `%wait` ?

rasciodc:
> 
> 
> have you tried `%wait` ?
> https://github.com/vlachoudis/bCNC/wiki/Scripting

Yes, I did, just like this:

    G91 G38.2 X-5 F100
    %wait
    %xleft=float(wx)
    G91 G38.2 X5 F100
    %wait
    %xright=float(wx)
    %center_x=0.5*(xleft+xright)
    G91 G0 X(center_x)
    G92 X[0]
 It did not work... =/

Harvie:
There seems to be some issue with %wait, see #1304 we probably need to fix this

rasciodc:
> 
> 
> There seems to be some issue with %wait, see #1304 we probably need to fix this

What is the status on this? Is there any update? (I'm genuinely curious, not trying to be pushy or demanding anything lol) 

I saw at least other 2 users trying to do exactly what I was intending and don't know if they succeeded. Cheers!

rgodin974:
Hello, did you find a solution with %wait ? I would want like to do this kind of macro too. Thanks

rasciodc:
> 
> 
> Hello, did you find a solution with %wait ? I would want like to do this kind of macro too. Thanks

I still haven't any solutions to that... Has anyone?

PR #1817, opened 2023-03-05
Fix user button interpolation

jlirochon:
# Actual behavior
- Because config parser is [created with default values](https://github.com/vlachoudis/bCNC/blob/master/bCNC/Utils.py#L143), [BasicInterpolation](https://docs.python.org/3/library/configparser.html#configparser.BasicInterpolation) is enabled, which means we can refer to config values when editing a user defined button.
- For example, if you define `slow = 5000` in the `[Buttons]` section, you could use `M3S%(slow)s` in the `command` and it would be interpolated to `M3S5000`. This is limited because you can't interpolate values from outside `[Buttons]` section, but this is how it's implemented actually.
- It means that if you want to include a `%` text somewhere, for example you want to name your button `Set laser power to 5%` you have to double the `%` like this : `Set laser power to 5%%` or **Python will raise an Exception**.
# Problem
When you open the button edition dialog, values populated from the config are interpoled. This means you lose your double `%%`. `Set laser power to 5%%` becomes `Set laser power to 5%` again, and you can't validate unless you manually double all the `%` again before clicking OK.
# Steps to reproduce
- Edit a button
- Name your button `Set laser power to 5%%` and click OK
- Edit the button again
- click OK
- and Exception is raised and the dialog is not closed
# Proposed fix
Load raw values from the config. This way the values remains exactly as there were typed in first place.
# Alternatives
- Totally disable interpolation by using `config = configparser.ConfigParser(interpolation=None)`. Maybe interpolation is an unneeded feature here ? It would allow to input `%` without escaping it.
# Opening
We could switch to [ExtendedInterpolation](https://docs.python.org/3/library/configparser.html#configparser.ExtendedInterpolation). This type of interpolation does not use `%` and **gives more power because it allows accessing other sections of the config**. For example you could use the value of `CNC.feedmax_z` in a command or whatever. The problem here is that this type of interpolation use `$` instead of `%`, and you would have to double all the `$`, which would be boring for existing configs. Anyway the proposed fix is still valid.

### DIFF
--- a/bCNC/Utils.py
+++ b/bCNC/Utils.py
@@ -752,30 +752,30 @@ class UserButton(Ribbon.LabelButton):
         tkExtra.Balloon.set(self, tooltip)
 
     # ----------------------------------------------------------------------
-    def name(self):
+    def name(self, raw=False):
         try:
-            return config.get("Buttons", f"name.{int(self.button)}")
+            return config.get("Buttons", f"name.{int(self.button)}", raw=raw)
         except Exception:
             return str(self.button)
 
     # ----------------------------------------------------------------------
-    def icon(self):
+    def icon(self, raw=False):
         try:
-            return config.get("Buttons", f"icon.{int(self.button)}")
+            return config.get("Buttons", f"icon.{int(self.button)}", raw=raw)
         except Exception:
             return None
 
     # ----------------------------------------------------------------------
-    def tooltip(self):
+    def tooltip(self, raw=False):
         try:
-            return config.get("Buttons", f"tooltip.{int(self.button)}")
+            return config.get("Buttons", f"tooltip.{int(self.button)}", raw=raw)
         except Exception:
             return ""
 
     # ----------------------------------------------------------------------
-    def command(self):
+    def command(self, raw=False):
         try:
-            return config.get("Buttons", f"command.{int(self.button)}")
+            return config.get("Buttons", f"command.{int(self.button)}", raw=raw)
         except Exception:
             return ""
 
@@ -863,15 +863,15 @@ class UserButtonDialog(Toplevel):
         Button(f, text=_("Ok"), command=self.ok).pack(side=RIGHT)
 
         # Set variables
-        self.name.insert(0, self.button.name())
-        self.tooltip.insert(0, self.button.tooltip())
-        icon = self.button.icon()
+        self.name.insert(0, self.button.name(raw=True))
+        self.tooltip.insert(0, self.button.tooltip(raw=True))
+        icon = self.button.icon(raw=True)
         if icon is None:
             self.iconCombo.set(UserButtonDialog.NONE)
         else:
             self.iconCombo.set(icon)
         self.icon["image"] = icons.get(icon, "")
-        self.command.insert("1.0", self.button.command())
+        self.command.insert("1.0", self.button.command(raw=True))
 
         # Wait action
         self.wait_visibility()


Harvie:
I beleive users were actualy reporting scripting to be broken.
https://github.com/vlachoudis/bCNC/wiki/Scripting

Maybe this might be related???

eg.: #1661 #1289 #1304

jlirochon:
Oh... Thx @Harvie I was not aware of this scripting feature !

After digging into the history, **I believe that scripting feature stopped working at some point when switching to Python3**. But some of the issues you linked seem more related to the moment the script is interpreted VS the moment the commands are executed by the CNC.

If we test outside of bCNC using the following `test.ini` :
```
[Buttons]
foo = %wait
```

**Here is Python2** :
```
Python 2.7.18 (default, Mar  5 2023, 17:48:22)
[GCC Apple LLVM 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import ConfigParser
>>> config = ConfigParser.ConfigParser()
>>> config.read("test.ini")
['test.ini']
>>> config.get("Buttons", "foo")
'%wait'
```

**Here is Python3 (like actual bCNC code)** :
```
Python 3.9.7 (default, Sep 25 2021, 12:02:50)
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import configparser
>>> config = configparser.ConfigParser()
>>> config.read("test.ini")
['test.ini']
>>> config.get("Buttons", "foo")
  File "<stdin>", line 1, in <module>
  File "/Users/julien/.asdf/installs/python/3.9.7/lib/python3.9/configparser.py", line 799, in get
    return self._interpolation.before_get(self, section, option, value,
  File "/Users/julien/.asdf/installs/python/3.9.7/lib/python3.9/configparser.py", line 395, in before_get
    self._interpolate_some(parser, option, L, value, section, defaults, 1)
  File "/Users/julien/.asdf/installs/python/3.9.7/lib/python3.9/configparser.py", line 442, in _interpolate_some
    raise InterpolationSyntaxError(
configparser.InterpolationSyntaxError: '%' must be followed by '%' or '(', found: '%wait'
```

**Here is a working Python3 example, where interpolation is disabled** :
```
Python 3.9.7 (default, Sep 25 2021, 12:02:50)
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import configparser
>>> config = configparser.ConfigParser(interpolation=None)
>>> config.read("test.ini")
['test.ini']
>>> config.get("Buttons", "foo")
'%wait'
```

**Adding `interpolation=None` seems to be the simplest fix to restore the initial scripting behavior. Interpolation of config variables won't be possible anymore.** I don't know if someone was using this, or if it was an accidental feature. Merging this PR is not required.

Otherwise (requires merging this PR) :
1. consider updating the docs to escape all the `%` in scripts (double `%%`). No new feature, no removed feature, scripting works again with some changes : `%wait` becomes `%%wait`
2. consider changing the scripting prefix to something else, say `@` or whatever... `%` was not a good choice from the start since this is the default interpolation character. Looks like Python2 was more indulgent but that was by luck. No new feature, no removed feature, scripting works again with some changes : `%wait` becomes `@wait`
3. consider changing the interpolation character to `$`. This would allow accessing all config variables from other sections like this : `${CNC:feedmax_z}` BUT this would require escaping all configs like `$30=0` to `$$30=0`. More powerful interpolation, no removed feature, scripting works again as is, requires changes to configs using `$`

jlirochon:
I'm closing this one. #1829 is the new proposal.

PR #1818, opened 2023-03-05
Update bCNC version

LittlePierre:


LittlePierre:
To be applied after #1819 please

LittlePierre:
update version in #1819

PR #1822, opened 2023-03-13
Involute gear

### LINKED ISSUE #1636: Gear
The gear plugins generates an error in the head of the teeth, an error that is accentuated if the number of teeth is increased.

LittlePierre:
Fix #1636 

Using https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/PartDesign/fcgear/involute.py

I used the involute.py file unchanged and put in and bCNC/lib directory.

Please check for licensing, I think it's OK but am not an expert of licensing...

in blue, the gear generated by Freecad involute.py library
in grey, the gear as generated by the old bCNC gear plugin
I think we can remove the old gear plugin that is less precise.

image
image

The plugin manages
- internal / external gear
- default parameters of a gear
- and user can specify some specific parameters as described below
image

Tests realized : 
- No regression in bCNC
- Exceptions are managed (for example if we try to draw a 2 teeth gear)
- Internal / External 
- Default parameters / specific parameters
- Everything seems to be OK


Harvie:
I am not questioning fact that there might be different kinds of gears and that some of them might not work properly. And to be honest i barely care about designing gears myself, i don't know anything about gears.

I am just saying that having two separate gear plugins is confusing for the users, therefore i think it would make sense to merge them to single plugin with unified GUI which would enable user to choose and cut any kind of gear they want.

LittlePierre:
Thanks Harvie
So I think that we should remove gear.py from plugins.
It generates approximative spur gears (involute gears), but it does just a basic linear interpolation of an involute gear. I think that this plugin becomes useless with the new one.
The new plugin generates the same spur "regular" gears, but with a much more precise interpolation. (blue curve vs grey curve in figure above) and with more features.

So I include in this PR a last commit : suppression of gear.py, replaced by involuteGear.py (commit to come in the next minutes)



source : https://en.wikipedia.org/wiki/Gear

_Spur gears or straight-cut gears are the simplest type of gear. They consist of a cylinder or disk with teeth projecting radially. Though the teeth are not straight-sided (but usually of special form to achieve a constant drive ratio, mainly [involute](https://en.wikipedia.org/wiki/Involute_gear) but less commonly [cycloidal](https://en.wikipedia.org/wiki/Cycloid_gear)), the edge of each tooth is straight and aligned parallel to the axis of rotation. These gears mesh together correctly only if fitted to parallel shafts.[[23]](https://en.wikipedia.org/wiki/Gear#cite_note-23) No axial thrust is created by the tooth loads. Spur gears are excellent at moderate speeds but tend to be noisy at high speeds.[[24]](https://en.wikipedia.org/wiki/Gear#cite_note-24)_

LittlePierre:
Trying to be more clear : 

Both gear and involuteGear plugins do the same job : draw regular spur gears.
- gear plugin does a rough linearization.
- involuteGear draws exactly the same gears, with a much sharper linearization, and additional features.

I see no reason why we should keep the two plugins or try merge them, since they do exactly the same job, involuteGear just does it sharper.

Harvie:
Thanks

Issue #1824, opened 2023-03-18
Manual Tool Change (TLO) doesn't work


SamsonBox:
Hi,
I tried the manual tool change (TLO) and I found out, that it doesn't work. In fact all the motions and calculation are right, but the TOL doesn't survive the `purgeController`  at all. The reason is the folloing:
After all motions are done, bCNC state is changed to `Idle` in the `controllerStateChange` function and `jobDone` is called which in turn calls `purgeController`  from the `serialIO`-Thread. The `purgeController` stores the  "G"-variables and the "TLO", resets the controller (via a realtime commad) a and tries restores the "G"-variables and the "TLO". But in the next run `serialIO`-loop the reset answer ("Grbl xxx") ist parsed and the `_stop` is set to `True`. Due to this flag the gcode-queue gets cleared (`emptyQueue()`) and the commands to restore the "G"-variables and the "TLO" are gone. Have I done something wrong? Or do you have any suggestions to overcome ths problem?

Thank you

rschell:
I have noticed these same issues over the past year.  There are a number of fixes IMO that need to be applied to get TLO and Tools to function as expected. You might give these changes try, [fix_TLO_and_Tool_handling](https://github.com/rschell/bCNC/tree/fix_TLO_and_Tool_handling)

MARIOBASZ:
I am going to try. I have always used M6 with TLO and no problems.

SamsonBox:
@rschell I have quickly looked over your changes. When I understand your proposel correctly, than you store the values in a file and reload them after the controller reset. I'm sure if that ist the way to go. I think the controller reset should wait until the reset is really over and send the commands after wards.

rschell:
My computer/controller on my machine spend 98% of its time in a development mode.  The operating system/bCNC/controller are rebooted quite often.  The state of the machine doesn't change during these reboots.  So my goal was to return bCNC/machine state to that prior to these resets.

The code snippets you maybe most interested in are in _GenericController.py (softReset and purgeController).  I spent a few days figuring out that g43.1 in the G return only indicates that a value has been set and is not sufficient to reset the state.

SamsonBox:
You are right, these changes are good. But I think the soft_reset should become an event to get it dispatched from the serialIO thread and wait till the reset-reply is received

rschell:
I'll take a look at your suggestion.

JipeB:
Is it necessary to purge the controller at the end of the program?

[...]

PR #1828, opened 2023-03-28
Feat/asyncio

SamsonBox:
This pull request soves issues #1824. It uses asyncio to handle the serial communication. This can be done as python 3.8 is mandatory for bCNC. Now the reset command is synchonized with the respons. Also the clear message ist synchonized.

### DIFF
--- a/bCNC/Sender.py
+++ b/bCNC/Sender.py
@@ -4,6 +4,8 @@
 #  Email: vvlachoudis@gmail.com
 #   Date: 17-Jun-2015
 
+import asyncio
+import functools
 import glob
 import os
 import re
@@ -67,7 +69,6 @@ STATECOLOR = {
     NOT_CONNECTED: "OrangeRed",
 }
 
-
 # =============================================================================
 # bCNC Sender class
 # =============================================================================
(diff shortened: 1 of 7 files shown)


Harvie:
To be honest, i don't really understand this async/await concept. Guess i will have to read something about it first :-)
Maybe other members of bCNC community will help me with the review in the meantime...

SamsonBox:
Actually this is pretty cool. Is a kind of cooperative scheduling in an eventloop within one thread. So the loop is started in the originally serialIo-Thread. within this loop tasks can be started and the can give back control to the schedular by the await key word. But take you time to review the changes.

Harvie:
OK, i did some research. And while i agree that this is kinda cool.
It also feels like it might significantly impact readability and debugability of the code, since execution of individual routines will now be interdependent to some extent.

While i am not completely opposed to the idea, i would like to see what are the gains when compared to "synchronous" code in our use-case and whether they acutally outweigh the increase in maintenance complexity.

Also we might consider if it would make sense to split bCNC into several separate threads.
For instance i can image separating GUI, Serial RX, TX and CAM subsystems, so they do not block each other.
I would not be as lightweight as asyncio, but might fix even more issues.

SamsonBox:
So the problem I found is metioned in issue #1824. The basic problem is, that the controller reset is send from the serialIO thread but can not be synchronized to the execution end (which is the reading the initial test the controller sends). So my basic idea was to have two indevidual tasks (a reading and writing task) running in one thread. So one can wait in the writing task for the response from the reading task. You are right, this can be also solved by running these two tasks in seperat threads, but I think that is a overkill for the serial communication as it is relatively slow. Using two thread will not reduce the higher maintenance complexity. The GUI is allread in the mainthread and the CAM subsystem can be moved to it's one thread in both cases.

PR #1829, opened 2023-03-30
Disable ConfigParser interpolation

jlirochon:
As discussed [here](https://github.com/vlachoudis/bCNC/pull/1817#issuecomment-1455155475), in bCNC the interpolation of config variables seems to be a side effect rather than a feature. Currently, the interpolation makes user buttons barely usable. Let's disable it for now.


Harvie:
Thanks

PR #1830, opened 2023-04-02
Jipe b patch 2

JipeB:
Related to thread #1827

### DIFF
--- a/bCNC/CNC.py
+++ b/bCNC/CNC.py
@@ -746,6 +746,8 @@ class CNC:
         "stepover": 40.0,
         "PRB": None,
         "TLO": 0.0,
+        "Zsensor": 0.0
+        "sensor": ""
         "version": "",
         "controller": "",
         "running": False,
--- a/bCNC/ProbePage.py
+++ b/bCNC/ProbePage.py
@@ -2223,11 +2223,12 @@ class ToolFrame(CNCRibbon.PageFrame):
         self.setProbeParams()
 
     # -----------------------------------------------------------------------
-    def updateTool(self):
+    def updateToolHeight(self):
         state = self.toolHeight.cget("state")
         self.toolHeight.config(state=NORMAL)
         self.toolHeight.set(CNC.vars["toolheight"])
         self.toolHeight.config(state=state)
+        self.event_generate("<<StateTool>>")
 
     # -----------------------------------------------------------------------
     def calibrate(self, event=None):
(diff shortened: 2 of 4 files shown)


Harvie:
Can you please provide summary of what these changes do?

JipeB:
The purpose of this PR is to improve the ergonomics of the interface and obtain correct operation of the tool change procedure (TLO)
**Improvement of the interface:** 
The Tab named "common" is not very explicit for a new user and contains an input field (TLO) which actually corresponds to the height of the measurement sensor.
I propose that this tab be renamed in a more explicit way. When I published this PR I called it Z-Sensor, but in practice I think the term Probe Parameters would be more correct.

**Correct operation of the TLO:**
When determining the part origin in Z, compensation for the height of the measuring sensor (touch plate or sensor) is obtained by  a G43.1 Z[sensor thickness], which is inappropriate. G10 L20 P[actual work space] Z[sensor thickness] should be used in this case.
See additional information: [(https://github.com/vlachoudis/bCNC/discussions/1775)] [(https://github.com/vlachoudis/bCNC/discussions/1827)]
JP

Harvie:
It would help a bit to have before/after screenshots of the GUI changes...

Also is the "sensor height" good way to name things? There are few issues i have with this:

- GRBL does call this TLO. Therefore it might be confusing for people who are familiar with GRBL and other related SW.
- Some people might be entering TLO manualy without even using any kind of sensor. therefore "sensor height" is even more confusing in these cases.

That said, i am curretly not convinced that this change is a good one.

JipeB:
image
image


Knowledge of the height of the sensor is necessary for all referencing since it is placed between the surface to be referenced and the stock.

> GRBL does call this TLO. Therefore it might be confusing for people who are familiar with GRBL and other related SW.

There is no ambiguity for grbl, TLO is a tool length correction, and nothing else. When G43.1 Zxxx is called at the time of tool change, grbl loads a register which contains the length of the new tool. The tool length value will be added to each Z movement. This is how all industrial machines have worked since the beginning. This is also how LinuxCNC works with the G43 Hxxx function. The difference between LinuxCNC and bCNC is the storage of tool lengths in a database, in the first case and the measurement of the TLO at each tool change, in the other case.

> Some people might be entering TLO manualy without even using any kind of sensor. therefore "sensor height" is even more confusing in these cases.

To manually enter the length of a tool, it would need to have been measured outside the machine, on a dedicated fixture, and placed back on the machine spindle in exactly the same position.
With the assembly with ERxxx type clamps which equip most amateur machines, it is mission impossible.
The calibration procedure, native to bCNC, when correctly executed, allows this measurement automatically.
Please read [https://github.com/vlachoudis/bCNC/wiki/Manual-tool-change-:-TLO-procédure](url)

> That said, i am curretly not convinced that this change is a good one.

More and more CNC users are using Fusion360 or FreeCAD to generate their GCode. This software allows the use of several tools while respecting NIST definitions. However, we must face the facts, the tool change method (TLO) proposed by bCNC does not work in its current state.
J.P.

Issue #1835, opened 2023-05-16
GRBL1.py override code not functioning correctly


rschell:
I noticed when I use the Override slider to temporary adjust either the Feed Rate or the Spindle speed.  The selected item, feed or spindle, continue to change beyond the desired set point.  Code appears continue to issue override strings of either 10 or 1 increment/decrements until the control is reset.  

The problem here is the GRBL1 code issues the appropriate override string but the override status variable is not adjusted, so GRBL1 continues to issue override strings.  Doesn't appear to be any tracking of the override commands issued.

My fix would be to adjust the override status variable "OvFeed" and "OvSpindle" by the appropriate amount in GRBL1, so GRBL1 will stop issuing overrides when the status variable equals the target value.  Requires 5 additional statements for each the Feed and Spindle Overrides. 

Anybody else had issues with the Feed and Spindle override controls?

rschell:
Found that the $10 setting on my grblHAL controller was not reporting the override values.  Changing the $10 value from 187 to 251 also fixed the issue.

PR #1836, opened 2023-05-17
fix grbl1 Feed and spindle logic updating status

rschell:
Without these changes, GBRL1.py continue to increment/decrement Feed and Spindle override commands until the controller limit is reached, rather than stopping at the sellected override amount.

See issue #1835.


rschell:
Changing $10 value from 187 to 251 resolved the issue as well.

PR #1839, opened 2023-06-07
Add flatpak manifest

cmp168:
This is my first flatpak packet, i test it on fedora-38, not all func, but only week have cnc and use any app for this.


Harvie:
Thanks

PR #1849, opened 2023-09-19
Vertical scrollbar to make small screens easier to use

batman-ua:
Vertical scrollbar for left side of the screen. Allows to use smalls screens like 1024x600 (Raspberry Pi)
Screenshot 2023-09-18 215515

### DIFF
--- a/bCNC/VScrollFrame.py
+++ b/bCNC/VScrollFrame.py
@@ -1,0 +1,39 @@
+from tkinter import (
+    VERTICAL,
+    NW,
+    E,
+    SE,
+    Y,
+    TOP,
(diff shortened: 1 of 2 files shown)


Harvie:
thanks

Harvie:
Do you think this might have caused issue #1885 ?

Harvie:
I've reverted this due to issue it had caused. Also in CAM and Editor this led to scrollbar to be duplicated, which is not a good UX:

image

batman-ua:
It might. Looks related :(

batman-ua:
I will take a look when I have time

PR #1855, opened 2023-10-04
Fix unused errorReport variable

dwrobel:
Fixes:
 Utils.py:213:9: F841 local variable 'errorReport'
                 is assigned to but never used.

### DIFF
--- a/bCNC/Utils.py
+++ b/bCNC/Utils.py
@@ -210,7 +210,7 @@ def loadConfiguration(systemOnly=False):
         config.read(iniSystem)
     else:
         config.read([iniSystem, iniUser])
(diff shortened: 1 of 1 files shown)


Harvie:
does the underscore mean the variable will be ignored by the test?
maybe we can remove that functionality completely, when nothing is using it. what's the point of calling that anyway?

dwrobel:
> does the underscore mean the variable will be ignored

Thanks to underscore we assign a value to an existing variable (not to the one which doesn't exist), see:
https://github.com/search?q=repo%3Avlachoudis%2FbCNC%20_errorReport&type=code.

Harvie:
thanks

PR #1856, opened 2023-10-04
Allow to launch flatpak directly from desktop environment

dwrobel:
- allows the flatpak bundle to be launched directly from the desktop environment by installing both .desktop and icon files,

- fixes issues in .desktop file: $ desktop-file-validate bCNC.desktop bCNC.desktop: error: (will be fatal in the future): value "bCNC.png" \ for key "Icon" in group "Desktop Entry" is an icon name with an \ extension, but there should be no extension as described in the \ Icon Theme Specification if the value is not an absolute path bCNC.desktop: warning: value "" for key "Path" in group \ "Desktop Entry" does not look like an absolute path,

- adds permission to flatpak bundle for accessing serial ports, cameras and home directory. As a result, after installing the bundle, it is fully usable without any extra manual tweaking.

- adds a hint how to build a flatpak bundle using multiple cores and ccache.


Harvie:
Thanks

PR #1879, opened 2024-02-15
New Probe tab to set workspace origin using probe

pavlot:
This pull-request contains new Probe tab to quickly change workspace origin and detect tool diameter with probe
 
image

This tab has two tools to simplify origin definition.

### Tool diameter

This part could be used to determinate tool diameter. 
Idea is simple: probe width is known. So first touch left probe side, remember position. Next jump over probe, touch right side. 
Tool diameter will be equal: right position - probe width - left position.

This tool allow to move gantry in 4 directions: NORTH (+Y), SOUTH(-Y), EAST(+X), WEST(-X). 
No difference which one is selected, they can be selected by convenience reasons. Main prerequisite -  probe should be touch not longer than 5mm from current position.

**Parameters**
**Direction** - where to move gantry to touch probe
**Probe width** - width of the probe. Gantry will be moved to this distance to find opposite side of probe
**Probe height** - Z distance required to jump over probe
**Tool bounds** - this parameter will add extra shift before go down on opposite side, to avoid wide tool touch probe top. It should be greater or equal than tool diameter 

In situation when probe cannot be placed close to mill, gantry has to be moved close to probe.

### Setting origin

After tool diameter is known, origin can be set for each axis. Tool just touch probe from any side, and this position +/- tool radius set as origin on that axis. +/- sign depends from which side probe was contacted. For Z axis: probe position + probe height

**Parameters**
**Tool diameter** - can be set manually, or automatically detected from previous step
**Rapid move (X, Y, Z)** - distance from current position, where gantry should be placed on given axis. Useful when new origin is far away from current position and moving with probe feed rate will take too long.
**Probe move (X, Y, Z)** - distance where probe is expected

This operation should be repeated for each axis

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ build
 
 flatpak/builddir
 flatpak/.flatpak-builder
+
+ptar
--- a/bCNC/CNC.py
+++ b/bCNC/CNC.py
@@ -682,6 +682,8 @@ class CNC:
         "prbx": 0.0,
         "prby": 0.0,
         "prbz": 0.0,
+        "prb_first_edge_pos":0.0,
+        "prb_second_edge_pos":0.0,
         "prbcmd": "G38.2",
         "prbfeed": 10.0,
         "errline": "",
--- a/bCNC/ProbePage.py
+++ b/bCNC/ProbePage.py
@@ -72,6 +72,13 @@ TOOL_POLICY = [
 
 TOOL_WAIT = [_("ONLY before probing"), _("BEFORE & AFTER probing")]
 
+MOVE_DIRECTION = [
+    _("NORTH"),
+    _("SOUTH"),
+    _("EAST"),
+    _("WEST")
+]
+
 CAMERA_LOCATION = {
     "Gantry": NONE,
     "Top-Left": NW,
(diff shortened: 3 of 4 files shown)


Harvie:
> NORTH (+Y), SOUTH(-Y), EAST(+X), WEST(-X).

Lets stick with Y+  Y-  X+  X-, i've never heard machinist using geography terms.

>  detect tool diameter with probe

What is the use case for this? Do they sell milling bits with unknown diameters nowadays?

pavlot:
> > NORTH (+Y), SOUTH(-Y), EAST(+X), WEST(-X).
> 
> Lets stick with Y+ Y- X+ X-, i've never heard machinist using geography terms.

no problem, I can fix it.

> > detect tool diameter with probe
> 
> What is the use case for this? Do they sell milling bits with unknown diameters nowadays?

In my case - I have a box of china corn mills, it contains 3 and 3.175 mills which I use to cut pcb edge. This .175 mm hard to measure on corn surface, but it is enough to cutoff ground or narrow wire on pcb. So I measure it on running spindle, which give exact size of cutting width. I have graphite probe which is connected to end of spindles rotor, so it is possible to check it with running tool. This is on 3018

Harvie:
It's rather hard to imagine spinning bit making electrical contact with graphite part without milling it away.

PR #1881, opened 2024-03-10
Fix Spirograph not working, gcd require int

HomineLudens:
Just notice the plugin was broken, probably due to changes in python


Harvie:
Thank you!

PR #1883, opened 2024-04-12
Box cover is optional

gcobos:
Allows to optionally create boxes without a cover. Meaning that it will skip the zigzag on the top edges of the box, and omit the top side.

Screenshot


Harvie:
Thank you

PR #1886, opened 2024-04-30
Corrected camera index reading from config when multiple webcams

muaddibus:
Corrected camera index reading from config when multiple webcams
It was opening device index 0, regardless saved setting.

#1666

### DIFF
--- a/bCNC/Camera.py
+++ b/bCNC/Camera.py
@@ -39,7 +39,7 @@ class Camera:
         if cv is None:
             return
         self.prefix = prefix
-        self.idx = Utils.getInt("Camera", prefix)
+        self.idx = Utils.getInt("Camera", "webcam")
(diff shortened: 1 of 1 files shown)


Harvie:
Thanks for noticing this. What was the "prefix" variable used for originaly?

muaddibus:
Didnt found it to be used anywhere.

Harvie:
> Didnt found it to be used anywhere.

What?

image

muaddibus:
I misunderstood you. I ment i didnt fount self.idx variable used except for setting video capture device number, which is originally always set to zero, and doesnt take into account webcam param in config.

StefanBruens:
This is incorrect, `prefix` may be either "aligncam" or "webcam".

The current code works correctly, though the changed index in the config only applies after a restart.

PR #1893, opened 2024-06-24
Get rid of temporary files for web pendant

StefanBruens:
Both OpenCV (for the camera) as well as PIL.Image (for the Canvas) can work with buffer like objects, which then can be immediately written out by the webserver code.


Harvie:
Nice idea. Thanks.

PR #1904, opened 2024-10-04
Update org.freedesktop.Platform to 24.08

dwrobel:
Update due to flatpack's complaints:
 Info: org.freedesktop.Platform is end-of-life,
 with reason: org.freedesktop.Platform 22.08 is no longer
 receiving fixes and security updates.
 Please update to a supported runtime version.


Harvie:
thanks

PR #1905, opened 2024-10-04
Update opencv-python to 4.5.5.64 and restrict numpy to 1.x

dwrobel:
- Update due to flatpack's complaints: WARNING: The candidate selected for download or install is a yanked version: 'opencv-python' candidate (version 4.5.5.62 at https://files.pythonhosted.org/packages/9d/98/36bfcbff30da27dd6922ed73ca7802c37d87f77daf4c569da3dcb87b4296/opencv_python-4.5.5.62-cp36-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (from https://pypi.org/simple/opencv-python/) (requires-python:>=3.6)) Reason for being yanked: deprecated, use 4.5.5.64

- Restrict numpy to not installing 2.x version (due to >=1.12 version specified in requirements.txt) which causes run-time error:
  RuntimeError: module compiled against ABI version 0x1000009
  but this version of numpy is 0x2000000

- Align installed versions between setup.py, requirements.txt, and io.github.bcnc.json files.


Harvie:
is bCNC codebase not compatible with latest numpy and opencv? i am not sure how hard it would be to upgrade completely...

dwrobel:
I don't know but I assumed this will be addressed by https://github.com/vlachoudis/bCNC/pull/1900. What I wanted to address here is to fix the current issue on `bCNC` which you can observe building either `flatpak` version or trying to install dependencies on the clean installation. You will get `numpy` 2.x being installed and observe the `RuntimeError` _(mentioned in the git message above)_ as a result `Camera` button is greyed _(nothing related to camera works)_.

Issue #1910, opened 2024-11-10
Can't run bCNC on Windows 11 with Python 3.13


Neil8978:
I installed Python 3.13 from Python.org on a new Windows 11 notebook, then installed bCNC using 'pip install --upgrade bCNC' as per instructions from 'https://pypi.org/project/bCNC/' The following results when I attempt to run bCNC:
c:\Users\Neil>python -m bCNC
================================================================================
WARNING: bCNC was recently ported to only support
python3.8 and newer.
Most things seem to work reasonably well.
Please report any issues to: https://github.com/vlachoudis/bCNC/issues
================================================================================
new-config bCNC <configparser.ConfigParser object at 0x00000210E1D797F0>
RuntimeError: module compiled against ABI version 0x1000009 but this version of numpy is 0x2000000
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\Neil\AppData\Roaming\Python\Python313\site-packages\bCNC\__main__.py", line 226, in <module>
    main()
    ~~~~^^
  File "C:\Users\Neil\AppData\Roaming\Python\Python313\site-packages\bCNC\__main__.py", line 153, in main
    application = bmain.Application(className=f"  {Utils.__prg__}  ")
  File "C:\Users\Neil\AppData\Roaming\Python\Python313\site-packages\bCNC\bmain.py", line 147, in __init__
    Sender.__init__(self)
    ~~~~~~~~~~~~~~~^^^^^^
  File "C:\Users\Neil\AppData\Roaming\Python\Python313\site-packages\bCNC\Sender.py", line 91, in __init__
    self.controllerLoad()
    ~~~~~~~~~~~~~~~~~~~^^
  File "C:\Users\Neil\AppData\Roaming\Python\Python313\site-packages\bCNC\Sender.py", line 134, in controllerLoad
    self.controllers[name] = eval(f"{name}.Controller(self)")
                             ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 1, in <module>
NameError: name 'G2Core' is not defined. Did you mean: 'GCode'?

c:\Users\Neil>

I could not see any previous reports of this error but there are some similarities to issue #1901

jpagliaccio:
I have the same problem on Linux (Fedora 41) solved it by also installing and running with Python 3.9. 

[...]

PR #1911, opened 2024-11-14
Fix dynamic loading of controllers and plugins in Python 3.13

rschell:
This should be a fix for #1910.  Tested on both 3.13 and 3.12 on Debian Bookworm


Harvie:
Thanks

PR #1913, opened 2024-11-19
Fix Display of abc coordinates

koneklaudio:
Fix indentation of several lines, in order to remove them from the exception


gcobos (approved):
@Harvie This seems to correct an existing bug in the code.

Harvie:
thanks

PR #1914, opened 2024-11-24
use correct NONE to compare cameraAnchor

tomjnixon:
Comparison was with None, but tk.NONE is used for this variable, which are different.

This broke the "gantry" camera mode for me; with this change it works as expected.

### DIFF
--- a/bCNC/CNCCanvas.py
+++ b/bCNC/CNCCanvas.py
@@ -1324,7 +1324,7 @@ class CNCCanvas(Canvas):
         hc //= 2
         x = w // 2  # everything on center
         y = h // 2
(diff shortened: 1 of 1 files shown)


tomjnixon:
Sure, that might be better, i just did it this way to be consistent with other uses of NONE in this file. I can go through and change those to use `tk.NONE` too, if you like? Thanks for having a look.

Harvie:
It seems it's more than that. it's multiple files and multiple constants like FALSE and TRUE. perhaps we should just leave it as is and be mindful of this....

Harvie:
Thanks

PR #1915, opened 2024-11-24
format numbers with a sensible number of digits when setting wcs

tomjnixon:
This mainly affected setting the work coordinate with the mouse, which can result in numbers with lots of digits. This doesn't work with my particular controller (uCNC), but more generally always results in a very long command being shown in the UI, and is inconsistent with other code.

This does make some uses of `_wcsSet` look a bit odd:

- some places pass '"0"' strings (or empty strings), presumably to avoid being formatted as "0.0", but this is redundant if they are always converted to a float first.
- some places round numbers before calling this, but that's not necessary.

I can fix these if desired. It's also unclear what the 10000mm limit is for; looks like this was introduced in this commit d8dbd633b70a1ded4f291df86d44bd527cd1c755.

### DIFF
--- a/bCNC/controllers/_GenericController.py
+++ b/bCNC/controllers/_GenericController.py
@@ -154,17 +154,17 @@ class _GenericController:
 
         pos = ""
         if x is not None and abs(float(x)) < 10000.0:
-            pos += "X" + str(x)
+            pos += CNC.fmt("X", float(x))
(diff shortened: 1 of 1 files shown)


Harvie:
> I can fix these if desired. It's also unclear what the 10000mm limit is for

Maybe @vlachoudis wanted to know if someone is crazy enough to run bCNC at CNC larger than 10 meters :-D

PR #1916, opened 2024-11-24
GRBL1: account for Hold sub-states in wait condition

tomjnixon:
This fixes a problem i had in `toolChange`, where `%wait` continues too soon after a command which transitions through an idle state, which isn't recognised if it's sent as `Hold:0` or `Hold:1`.

As per my other PR, this caused an issue with uCNC, and might not with other controllers, though GRBL does apparently send `Hold:0` and `Hold:1`.


Harvie:
Thanks

PR #1922, opened 2025-01-28
Add NixOS (nix-shell) support

nuxeh:
Add a shell which facilitates development and running of bCNC on a NixOS system, or another *nix using the Nix package manager.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,13 @@
+let
+  pkgs = import <nixpkgs> {};
+in pkgs.mkShell {
+  packages = [
+    (pkgs.python3.withPackages (python-pkgs: [
+      python-pkgs.tkinter
+      python-pkgs.numpy
+      python-pkgs.svgelements
+      python-pkgs.pyserial
+      python-pkgs.pillow
+    ]))
+  ];
+}


Harvie:
I appreciate such efforts, but i am not very knowledgable about nix os. how exactly would this be used? (maybe couple sentences in README.md would be useful?). Also, does this have to be in repo root?

nuxeh:
Thanks for getting back so quickly. If you have a system with the Nix package manager installed, simply `cd` into the project's source directory, where `shell.nix` resides, and running `nix-shell` will source all dependencies - in this case - python, and the required python dependencies to run or develop on bCNC, and drop you into an ephemeral shell environment containing these dependencies, while not touching your system. It's pretty handy, and especially if you use NixOS, it's the easiest way by far to run something that's not already included in NixOS's package repository.

Regarding documentation, as you can see above it's pretty trivial, and if you already use nix you are not likely to need any documentation. If you don't use Nix, then chances are you probably don't care. I could however note the basic nix support in the readme if you like. Having a shell is only the first step in full nix support, adding a "flake" for example (too much to go into here) would be another, single file called `flake.nix`, which would conventionally be in the root. As an aside, I've actually been thinking of packaging bCNC for NixOS for a while, but never got around to it, though that wouldn't touch bCNC's repo.

It doesn't _have_ to be in the repo root, but this is the usual place for it to be. Also, in a sense, for people who use nixos, when looking at a given software project, the presence of a `shell.nix` in the project root is a quick and sure indication that the project has some nix support. In a sense the presence of nix files in the root advertises nix support for those who are looking for it. Besides, where else would you rather put it? The only thing that comes to mind is a 'nix' directory, but again, this isn't the usual way to do  it, and it would only contain one file, so the amount of clutter is similar, I feel.

Harvie:
> As an aside, I've actually been thinking of packaging bCNC for NixOS for a while, but never got around to it, though that wouldn't touch bCNC's repo.

I have mixed feelings about those... For example:
As ArchLinux user i love the idea of having bCNC ArchLinux package, because it's easy to install. As bCNC developer, i hate idea of having to update any distro-specific package with every release, because there are lots of distributions and it seems to add up rather quickly. So i feel like pip install is gonna be only officaly supported way of installing bCNC, since it seems to work everywhere without major issues. But i don't really mind this stuff as long as i am not the one who has to maintain it, so i am merging, but don't expect me to fix it if something goes wrong in future. That's your job now :-P

nuxeh:
I think it's good to have packages for as many platforms as possible for wider accessibility.

But true story, Nix packages can very simply be made to self update, such that code changes upstream, a PR is made on the nixpkgs repo for review. So I don't mind at all

Thanks for the merge!

PR #1923, opened 2025-01-28
Update python-opencv version

hannesweisbach:
python-opencv 4.5.5.62 has been yanked and is no longer available: https://pypi.org/project/opencv-python/4.5.5.62/

4.5.5.64 is the next version up that is still available, but not the newest version. Maybe there is interest in generally updating; the current version of python-opencv is 4.1.0.68. 

However I don't understand the implications of updating, so I stuck with recommended version.


Harvie:
thanks

PR #1937, opened 2025-05-15
Fix: module 'PIL.Image' has no attribute 'ANTIALIAS' in sketcher and pyrograph

keesj:
Following the 10.0.0 release of PIL it no longer has the ANTIALIAS  constant hence updated the references

https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html#constants

Also see
[Stackoverfow](https://stackoverflow.com/questions/76616042/attributeerror-module-pil-image-has-no-attribute-antialias)


Harvie:
Thanks

PR #1938, opened 2025-05-19
text-plugin: Remove separation marks between characters when processing SHX fonts

gcobos:
I have noticed that when using SHX fonts there's a dot drilled in between every letter. It's not a problem with the font itself, and it happens always. I have modified the code to delegate the `CNC.zenter(depth)` command to where it's actually used (before a `gline`), instead of doing it at the end of every `grapid` movement. This saves lots of time and removes the bug of the extra drill between letters.

### DIFF
--- a/bCNC/plugins/text.py
+++ b/bCNC/plugins/text.py
@@ -8,6 +8,7 @@ from ToolsPage import Plugin
 from PIL.FontFile import FontFile
 from shxparser.shxparser import ShxFont,ShxPath
 from svgelements import Arc
+
 __author__ = "Filippo Rivato"
 __email__ = "f.rivato@gmail.com"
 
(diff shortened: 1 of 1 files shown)


Harvie:
Hi. What is separation mark? Do you have screenshot?

gcobos:
I am still working on the solution, but the problem is that a dot is drilled between every letter.
Screenshot from 2025-05-19 21-33-33
Screenshot from 2025-05-19 21-52-08

gcobos:
Here, the result of the job. Not the drills between every letter. I am trying to avoid those.

Harvie:
> I am still working on the solution

Ok thanks. Let me know when it's ready and i will merge.

gcobos:
Now I'm happy with the result. This makes the jobs much faster
Screenshot from 2025-05-20 23-48-35

gcobos:
@Harvie Please, merge this PR

Harvie:
Thank you

Issue #1953, opened 2025-08-12
Check for updates no longer works


tpimh:
It still checks for latest release on GitHub, and tries to compare it to the currently running version:

```
Traceback (most recent call last):
  File "/home/dima/Documents/appimage/bcnc/bCNC/bCNC.AppDir/usr/lib/python3.10/site-packages/bCNC/Utils.py", line 533, in __call__
    return self.func(*args)
  File "/home/dima/Documents/appimage/bcnc/bCNC/bCNC.AppDir/usr/lib/python3.10/site-packages/bCNC/Updates.py", line 174, in check
    if self.isNewer(latest_version):
  File "/home/dima/Documents/appimage/bcnc/bCNC/bCNC.AppDir/usr/lib/python3.10/site-packages/bCNC/Updates.py", line 152, in isNewer
    for a, b in zip(av, bv):
ValueError: invalid literal for int() with base 10: 'pypi'
```

Should check for the [latest release on PyPI](https://pypi.org/project/bCNC/#history) instead.

PR #1963, opened 2025-09-30
Use pypi to check for updates

MarkHerhold:
Check updates from PyPI and updates UI release message. Solves #1953

This is my first PR to this repo. Comments/suggestions welcome.

### DIFF
--- a/bCNC/Updates.py
+++ b/bCNC/Updates.py
@@ -5,6 +5,7 @@
 
 import json
 import time
+from datetime import datetime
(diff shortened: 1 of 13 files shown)


Harvie:
Great idea, thanks! Just a tiny question: what happens if the pypi is not accessible? isn't this gonna affect offline/firewalled users? it would be shame if pypi downtime caused localy installed software to fail (i know you just reused existing code, but i am not sure what it was doing before...)

PR #1969, opened 2025-12-13
added support for grbl 1.1f found on Genmitsu CNC

YAMLcase:


YAMLcase:
interesting, I wonder why manufacturers would decide to take the hard path instead of just use grbl.  regardless, this is just a draft and no hard feelings if you want me to call it something else (fakegrbl, grblmitsu, etc) or reject it outright.

Harvie:
> if you want me to call it something else

- i am ok with calling it something like grbl-genmitsu
- first we need to understand what the changes actualy are, are you sure it's really different? are you aware that GRBL has options to configure the protocol compatibility level? it needs to be correctly configured to work with bCNC.
- have you considered using upstream grbl? often times it's best way in all regards. what the MCU us in genmitsu?

dukicn:
I am having the same issue with Genmitsu 3020 Pro Max v2. I just switched to bCNC from UGS due to autoleveling capabilities and every time I do certain actions like autoleveling or even just pressing stop when joging the axes, bCNC and CNC loose communication until I either perform the hard reset or disconnect and connect. I managed to fix the error myself and just wanted to create a pull request when I saw this one so no point in creating another one. Will this be merged to main?

Issue #1982, opened 2026-05-01
CAM/Profile adding segments


slazav:
Could be related with #1981.
Extra circle appears when doing Profile, see screenshot:
<img width="1065" height="666" alt="Image" src="https://github.com/user-attachments/assets/35ba83a2-5da8-4aeb-ab3b-28cb65f0f1d0" />
Test DXF file, to be used with D2.0 mill: [cut_test2.dxf.txt](https://github.com/user-attachments/files/27271381/cut_test2.dxf.txt)

MARIOBASZ:
Hello guys.

As others have mentioned, the problem arises from the threshold value used to remove the minimum segments, specifically in the arcs. I used `linearize` in the original figure, and the generated profile is correct.

Has any problem been reported if the figure only uses G1?

The problem was identical with both `Profile` and `Offset`.

Solution found:

I changed the sequence for removing the minimum segments:

If `removeZerolength(abs(offset)/100)` is executed before to `.intersectself()` in both the `endmilloffset.py` plugin and `offsetClean` (in `bpat.py`), the trajectories seem to be generated without problems.

MARIOBASZ:
There are still problems and incorrect geometries in other examples. If the shortest segments are not removed, are errors reported? Tomorrow I will follow up on this idea: If the error is generated by errors in curved segments (start point - end point - determined radius), removing the shortest segments should meet the condition:  if the preceding and following segments are lines.

MARIOBASZ:
This solution of checking whether the next segment is not a line segment  seems consistent. So far, the results have been correct.
Problem #1981 appears to have a different origin.

MARIOBASZ:
I'm uploading sample examples where errors occur.
Remove the .TXT extension for testing purposes. The name indicates the operation causing the problem

[cut_test3.BIT_0.20_OUT.ngc.TXT](https://github.com/user-attachments/files/28032124/cut_test3.BIT_0.20_OUT.ngc.TXT)
[REST MACHINING_4.BIT_2.00_OUT.ngc.TXT](https://github.com/user-attachments/files/28032126/REST.MACHINING_4.BIT_2.00_OUT.ngc.TXT)
[REST MACHINING_6.BIT_2-00_OUT.ngc.TXT](https://github.com/user-attachments/files/28032122/REST.MACHINING_6.BIT_2-00_OUT.ngc.TXT)
[...]

Issue #1981, opened 2026-05-01
CAM/Profile losing segments


slazav:
For certain shapes CAM/Profile produces wrong result (see screenshot). I'm using recent version 0.9.16. I think, it happens when a perpendicular segment with length = mill radius appears. For this test D=0.2mm mill is used.
<img width="1065" height="666" alt="Image" src="https://github.com/user-attachments/assets/35de849b-c8e8-4e9e-92a8-c4fd3369b98c" />

slazav:
Test DXF file [cut_test.dxf.txt](https://github.com/user-attachments/files/27271344/cut_test.dxf.txt), to be used with D0.2mm mill

MARIOBASZ:
I believe the problem arises from the following: `intersectSelf()` correctly generates the segments. The error seems to be in the assignment of `mark as _cross for later evaluation` in `removeExclude()`.

PR #1983, opened 2026-05-19
Problems generating Profile

MARIOBASZ:
Issues #1981 #1982


MARIOBASZ:
I believe IntersectPath should be fixed in the same way IntersectSelf was fixed.

MARIOBASZ:
@Harvie, don't commit to this yet. Some of the changes made are causing Pocket to freeze again (an old problem). It didn't happen in my version, which was disconnected from GitHub. I'll check it out.

